PR #11, opened 2019-02-15
Access metadata using optics

### LINKED ISSUE #4: Access metadata using optics
Library design assumes usage of `Lens` and `Prism` for metadata access (see #2). Since one of the goals of `scalaz-distributed` is to minimize the number of external dependencies, it is not an option to use `Monocle`. 

Provide an implementation of `Lens` and `Prism`. In case additional typeclasses are introduced, briefly explain their purpose in the context of `scalaz-distributed`.

pshemass:
Closes #4.

### DIFF
--- a/src/main/scala/scalaz/DistributedModule.scala
+++ b/src/main/scala/scalaz/DistributedModule.scala
@@ -9,9 +9,13 @@ trait Member
 sealed trait Membership
 
 object Membership {
-  final case class Join(member: Member)        extends Membership
-  final case class Leave(member: Member)       extends Membership
+
+  final case class Join(member: Member) extends Membership
+
+  final case class Leave(member: Member) extends Membership
+
   final case class Unreachable(member: Member) extends Membership
+
 }
 
 sealed trait DistributedError
@@ -19,14 +23,19 @@ sealed trait DistributedError
 case class MetadataID(v: String)
 
 trait DistributedModule {
-
   type F[A] = IO[DistributedError, A]
 
   type Type[A]
 
-  type Lens[A, B]
-  type Prism[A, B]
-  type Traversal[A, B]
+  type Path[A, B]
+
+  implicit class PathSyntax[A: Type, B: Type](self: Path[A, B]) {
+    def >>> [C: Type](that: Path[B, C]): Path[A, C] = compose[A, B, C](self, that)
+  }
+
+  def key[K: Type, V: Type](k: K): Path[Map[K, V], V]
+
+  def compose[A: Type, B: Type, C: Type](l: Path[A, B], r: Path[B, C]): Path[A, C]
 
   implicit val stringType: Type[String]
   implicit val longType: Type[Long]
@@ -43,10 +52,17 @@ trait DistributedModule {
   }
 
   trait Metadata[A] {
-    def set[B: Type](where: Lens[A, B], b: B)(implicit A: Type[A]): F[Unit]
-    def get[B: Type](where: Lens[A, B])(implicit A: Type[A]): F[B]
-    def increment(where: Lens[A, Int])(implicit A: Type[A]): F[Unit]
+    def set[B: Type](where: Path[A, B], b: B)(implicit A: Type[A]): F[Unit]
+
+    final def modify[B: Type](where: Path[A, B], f: B => B)(implicit A: Type[A]): F[Unit] =
+      for {
+        c <- get[B](where)
+        r <- set(where, f(c))
+      } yield r
+
+    def get[B: Type](where: Path[A, B])(implicit A: Type[A]): F[B]
+
+    def increment(where: Path[A, Int])(implicit A: Type[A]): F[Unit]
   }
 
-  def key[K, V](k: K): Lens[Map[K, V], V]
 }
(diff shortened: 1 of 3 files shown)


src/main/scala/scalaz/DistributedModule.scala
```diff
@@ -9,24 +9,44 @@ trait Member
...
-  type Lens[A, B]
-  type Prism[A, B]
-  type Traversal[A, B]
+  type Path[A, B] = PathElem[A, B]
+
+  sealed trait PathElem[A, B]
+
+  private object PathElem {
+
+    case class Key[K: Type, V: Type](v: K) extends PathElem[Map[K, V], V]
```
mijicd: Please also finalize these.

src/main/scala/scalaz/DistributedModule.scala
```diff
@@ -9,24 +9,44 @@ trait Member
...
+
+    case class Elements[A: Type]() extends PathElem[Set[A], A]
+
+    case class Composed[A: Type, B: Type, C: Type](l: PathElem[A, B], r: PathElem[B, C])
+        extends PathElem[A, C]
+
+  }
+
+  implicit class LensSyntax[A: Type, B: Type](self: Path[A, B]) {
```
mijicd: Please rename to `PathSyntax`.

jdegoes:
As was originally sketched out, you have to _completely separate_ module definition from implementation. The distributed module will define the _structure_ but leave many types _abstract_. That's the only way to do this.

Roughly:

```scala
trait DistributedModule {
   type Path[A, B]
  ...
}
private[scalaz] class DistributedModuleImpl {
}
package object scalaz {
  val distributed: DistributedModule = new DistributedModuleImpl { }
}
```

You cannot implement _anything_ in the module definition. The types are abstract _by design_ to prevent leaking implementation details and allow reifying types and dropping type parameters in the actual implementation.

src/main/scala/scalaz/DistributedModule.scala
pshemass: what do mean by that?

src/main/scala/scalaz/DistributedModule.scala
pshemass: done

src/main/scala/scalaz/DistributedModule.scala
mijicd: Just make them `final`.

src/main/scala/scalaz/DistributedModule.scala
```diff
@@ -43,10 +52,17 @@ trait DistributedModule {
   }
 
   trait Metadata[A] {
-    def set[B: Type](where: Lens[A, B], b: B)(implicit A: Type[A]): F[Unit]
-    def get[B: Type](where: Lens[A, B])(implicit A: Type[A]): F[B]
-    def increment(where: Lens[A, Int])(implicit A: Type[A]): F[Unit]
+    def set[B: Type](where: Path[A, B], b: B)(implicit A: Type[A]): F[Unit]
+
+    def modify[B: Type](where: Path[A, B], f: B => B)(implicit A: Type[A]): F[Unit] =
```
mijicd: I think `modify` can be `final` as well.

src/main/scala/scalaz/DistributedModuleImpl.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  sealed class TypeImpl[A]
+
+  type Type[A] = TypeImpl[A]
+
+  sealed trait PathElem[A, B]
+
+  private object PathElem {
+
+    case class Key[K: Type, V: Type](v: K) extends PathElem[Map[K, V], V]
```
mijicd: See the previous conversation.

src/main/scala/scalaz/DistributedModule.scala
pshemass: done

src/main/scala/scalaz/DistributedModuleImpl.scala
pshemass: done

src/main/scala/scalaz/DistributedModule.scala
pshemass: understood. done

src/main/scala/scalaz/DistributedModule.scala
pshemass: done

src/main/scala/scalaz/DistributedModule.scala
pshemass: sorry but this will cause: Error: The outer reference in this type test cannot be checked at run time.

src/main/scala/scalaz/DistributedModule.scala
pshemass: done

src/main/scala/scalaz/DistributedModuleImpl.scala
pshemass: see my previous response

mijicd:
@pshemass nice work, thanks!

PR #12, opened 2019-03-19
Upgrade ZIO version

mijicd:

PR #13, opened 2019-03-19
Restructure distributed module

mijicd:
### Summary

- Renamed `DistributedModule` to `Client`
- Split file into separate package-level artifacts
- Removed `DistributedModuleImpl` 

### Remarks

Branched from #12 .


build.sbt
```diff
@@ -1,15 +1,12 @@
...
-)
+ThisBuild / scalaVersion := "2.12.8"
+ThisBuild / version := "0.1.0-SNAPSHOT"
 
 lazy val root = project
   .in(file("."))
   .settings(
-    libraryDependencies += "org.scalaz" %% "scalaz-zio" % "0.6.0"
+    name := "scalaz-distributed",
+    libraryDependencies += "org.scalaz" %% "scalaz-zio" % "1.0-RC1"
```
jdegoes: Same here, I'd depend on `0.16` or higher, not the incompletely-deployed RC version.

src/main/scala/scalaz/distributed/Client.scala
```diff
@@ -0,0 +1,34 @@
...
+      type Path[A, B] = PathElem[A, B]
+
+      override def key[K: Type, V: Type](k: K): Path[Map[K, V], V] = PathElem.Key[K, V](k)
+
+      override def compose[A: Type, B: Type, C: Type](x: Path[A, B], y: Path[B, C]): Path[A, C] =
+        x >>> y
+
+      override def connect(member: Member, seed: Set[InetAddress]): F[Protocol[Type, Path]] = ???
+
+      override def members(cb: Membership => F[Boolean]): F[Unit] = ???
```
jdegoes: We can think about changing this now that we have ZIO Stream, e.g.:

```scala
def members: Managed[Stream[Nothing, Membership]]
```

Actually we need something that's a combination of a stream + "current value", like `Signal`, which holds all the members and updates that over time. That would be more convenient to use. But we can always change that later.

src/main/scala/scalaz/distributed/package.scala
```diff
@@ -0,0 +1,7 @@
+package scalaz
+
+import scalaz.zio.IO
+
+package object distributed {
+  type F[A] = IO[Error, A]
```
jdegoes: I'd name this more helpfully, maybe `Distributed[A] = IO[Error, A]`

src/main/scala/scalaz/distributed/SupportedType.scala
```diff
@@ -0,0 +1,15 @@
+package scalaz.distributed
+
+private[distributed] trait SupportedType[A]
```
jdegoes: We'll need to reify the type here, add some method like `reifiedType` that returns a sum type, of all possible types.

src/main/scala/scalaz/distributed/PathElem.scala
```diff
@@ -0,0 +1,18 @@
...
+  final case class Key[K: SupportedType, V: SupportedType](v: K) extends PathElem[Map[K, V], V]
+
+  final case class Elements[A: SupportedType]() extends PathElem[Set[A], A]
+
+  final case class Composed[A: SupportedType, B: SupportedType, C: SupportedType](
+    x: PathElem[A, B],
+    y: PathElem[B, C]
+  ) extends PathElem[A, C]
+
+  implicit class PathSyntax[A: SupportedType, B: SupportedType](self: PathElem[A, B]) {
```
jdegoes: The syntax should probably not be in a private class but closer to the top, maybe package object.

src/main/scala/scalaz/distributed/Protocol.scala
```diff
@@ -0,0 +1,5 @@
+package scalaz.distributed
+
+trait Protocol[Type[_], Path[_, _]] {
```
jdegoes: If you push this back in the module, you can get rid of the type parameters, but it's a personal preference. It may just be clunky to use since the user will never know the actual type of `Type` or `Path`.

src/main/scala/scalaz/distributed/Metadata.scala
```diff
@@ -0,0 +1,14 @@
+package scalaz.distributed
+
+trait Metadata[Type[_], Path[_, _]] {
+  def get[A: Type, B: Type](where: Path[A, B]): F[B]
+  def set[A: Type, B: Type](where: Path[A, B], b: B): F[Unit]
+
+  final def modify[A: Type, B: Type](where: Path[A, B], f: B => B): F[Unit] =
```
jdegoes: This won't be atomic... instead of `modify` we may have to introduce type-specific operations based on the implementation. e.g. for an `Int` we can do atomic increment / decrement using CRDTs.

src/main/scala/scalaz/distributed/Protocol.scala
mijicd: Moved both `Protocol` and `Metadata` back to `Client`.

src/main/scala/scalaz/distributed/package.scala
mijicd: Done.

build.sbt
mijicd: This is due to rebasing on #12.

Issue #8, opened 2019-02-11
Create a microsite


mijicd:
Document and publish a microsite using [mdoc](https://github.com/scalameta/mdoc). Use ZIO's own documentation as reference.

PR #22, opened 2019-04-06
Write README

### LINKED ISSUE #20: Document the library
Write a clear statement of purpose and provide a high-level architecture overview. Enlist all of the resources (videos, papers, books) as part of the "getting started" guide.

mijicd:
Closes #20.

### Summary

- Briefly described the library and enlisted necessary background readings in readme.
- Added contributing guide.
- Added code of conduct.

### Todo

I propose to expand readme and add microsite in a separate issue (#8), as soon as we polish all the details as we spoke in the sync (i.e. which CRDTs to support, basic integration work etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ of the library. For an introduction to distributed systems in general, make sure
 Foundations:
 
 - [Conflict-free Replicated Data Types][link-crdts-paper]
-- [CRDTs illustrated][link-crds-illustrated]
+- [CRDTs illustrated][link-crdts-illustrated]
 - [Strong Eventual Consistency and Conflict-free Replicated Data Types][link-shapiro]
 
 Production experiences:
--- a/README.md
+++ b/README.md
@@ -1,3 +1,60 @@
-# ZIO keeper
+# Ziokeeper
 
-[![Gitter](https://badges.gitter.im/scalaz/scalaz-distributed.svg)](https://gitter.im/scalaz/scalaz-distributed?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,3 +1,60 @@
-# ZIO keeper
+# Ziokeeper
 
-[![Gitter](https://badges.gitter.im/scalaz/scalaz-distributed.svg)](https://gitter.im/scalaz/scalaz-distributed?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![TravisCI][badge-travis]][link-travis]
+[![Gitter][badge-gitter]][link-gitter]
+
+Ziokeeper is a purely-functional, type-safe library for managing distributed state in a strong
```
pshemass: I would try to describe Ziokeeper as something like lightweight version of etcd(because Zookeeper is a little bit more than metadata store) and CRDT as first version of implementation. 

> Ziokeeper is a purely-functional, type-safe library for managing in-memory key-value state across nodes in cluster. First approach to achieve consistency in the cluster we plan to use CRDT. 

or something along those lines. Then I would set expectations on right level by listing functionality that we want to implement.

- cluster formation and management via gossip
- be able to subscribe on cluster events - join,leave
- access to value by key - probably we should list types that we are supporting as elements in maps right now 
- be able to subscribe on changes on particular value - modify, remove

I would also mentioned something about transport protocol(UDP/TCP) and serialization

mijicd: I've enumerated the project goals as suggested. Didn't want to provide much details about the implementation as we still have a lot of "rough" edges, but the points you made here should make the purpose of the library clear(er).

pshemass: agree

PR #24, opened 2019-04-15
Feature/crdt or set

pshemass:
first implementation of of ORSet and Vector Clock

### DIFF
--- a/src/main/scala/scalaz/ziokeeper/VectorClock.scala
+++ b/src/main/scala/scalaz/ziokeeper/VectorClock.scala
@@ -1,0 +1,101 @@
+package scalaz.ziokeeper
+
+import scala.annotation.tailrec
+import VectorClock._
+import scalaz.ziokeeper.VectorClock.VectorClockOrdering.{ After, Before, Concurrent, TheSame }
+
+/**
+ * Implements an algorithm for generating a partial ordering of events in a distributed system and detecting causality violations.
+ *
+ * In the time-stamping system, each process has a VectorClock that is maintained as follows:
+ * - increment local counter in order to track progress whatever performs actions(e.g. add, remove, modify)
+ * as well as when you receive and send messages from/to other members of the cluster.
+ * - when a cluster member sends a message, it attaches to it the current value of VectorClock
+ * - whenever we receive event from other cluster members we takes max between position stored in local vector clock about positions of others nodes and
+ *  respective positions from incoming message.
+ *
+ * NOTE: Keep in mind that vector clock provides partial order not total order so there will still appear
+ * situations that two events in the distributed system with Vector Clock timestamping be concurrent to each other.
+ *
+ */
+case class VectorClock[A](value: Map[A, Long]) extends AnyVal {
+  def time(mid: A): Long = value.getOrElse(mid, 0L)
+
+  def increment(mid: A): VectorClock[A] =
+    VectorClock(this.value.updated(mid, time(mid) + 1))
+
+  def compare(right: VectorClock[A]): VectorClockOrdering = {
+    val leftWithDefault  = this.value.withDefaultValue(0L)
+    val rightWithDefault = right.value.withDefaultValue(0L)
+
+    @tailrec
+    def loop(keys: List[A], leftBigger: Boolean, rightBigger: Boolean): (Boolean, Boolean) =
+      keys match {
+        case _ if leftBigger && rightBigger => (leftBigger, rightBigger)
+        case Nil                            => (leftBigger, rightBigger)
+        case h :: t =>
+          val diff = leftWithDefault(h) - rightWithDefault(h)
+          loop(t, leftBigger || diff > 0, rightBigger || diff < 0)
+      }
+
+    val keys = leftWithDefault.keySet.union(rightWithDefault.keySet)
+
+    loop(keys.toList, leftBigger = false, rightBigger = false) match {
+      case (false, false) => TheSame
+      case (true, true)   => Concurrent
+      case (true, false)  => After
+      case (false, true)  => Before
+    }
+  }
+
+  def mergeWith(right: VectorClock[A])(f: (Long, Long) => Long): VectorClock[A] = {
+    val leftWithDefault  = this.value.withDefaultValue(0L)
+    val rightWithDefault = right.value.withDefaultValue(0L)
+
+    VectorClock(
+      leftWithDefault.keySet
+        .union(rightWithDefault.keySet)
+        .foldLeft(Map.empty[A, Long]) {
+          case (acc, key) =>
+            acc + (key -> f(leftWithDefault(key), rightWithDefault(key)))
+        }
+    )
+  }
+
+  def merge(right: VectorClock[A]): VectorClock[A] =
+    mergeWith(right)(_.max(_))
+}
+
+object VectorClock {
+  def empty[A] = VectorClock[A](value = Map.empty)
+
+  /**
+   * Represents Partial Ordering of the Vector clock.
+   *
+   * When we have cluster with three member with ids 111, 222, 333.
+   * Vector clocks initially will look like below on all nodes.
+   * (111 -> 0, 222 -> 0, 333 -> 0)
+   *
+   * At this point the are all TheSame.
+   *
+   * imagine that 111 and 222 perform add action
+   *
+   * 111 = (111 -> 1, 222 -> 0, 333 -> 0)
+   * 222 = (111 -> 0, 222 -> 1, 333 -> 0)
+   * 333 = (111 -> 0, 222 -> 0, 333 -> 0)
+   *
+   * 111 and 222 are Concurrent because you cannot say which happen first.
+   *
+   */
+  sealed trait VectorClockOrdering
+
+  object VectorClockOrdering {
+    case object TheSame extends VectorClockOrdering
+
+    case object Before extends VectorClockOrdering
+
+    case object After extends VectorClockOrdering
+
+    case object Concurrent extends VectorClockOrdering
+  }
+}
--- a/src/main/scala/scalaz/ziokeeper/crdt/ORSet.scala
+++ b/src/main/scala/scalaz/ziokeeper/crdt/ORSet.scala
@@ -1,0 +1,31 @@
+package scalaz.ziokeeper.crdt
+
+import scalaz.ziokeeper.VectorClock
+
+case class ORSet[A, B](values: Vector[ORSet.VersionedEntry[A, B]]) {
+
+  def localRemove(elem: A): (Vector[VectorClock[B]], ORSet[A, B]) = {
+    val (filtered, removedVcs) = this.values.foldLeft(
+      (Vector.empty[ORSet.VersionedEntry[A, B]], Vector.empty[VectorClock[B]])
+    ) {
+      case ((filteredAcc, vcs), ORSet.VersionedEntry(`elem`, c)) => (filteredAcc, vcs :+ c)
+      case ((filteredAcc, vcs), entry)                           => (filteredAcc :+ entry, vcs)
+    }
+    (removedVcs, ORSet[A, B](filtered))
+  }
+
+  def remove(elem: A, vcs: Vector[VectorClock[B]]): ORSet[A, B] =
+    ORSet(this.values.filterNot {
+      case ORSet.VersionedEntry(`elem`, v) => vcs.contains(v)
+      case _                               => false
+    })
+
+  def add(value: A, vectorClock: VectorClock[B]): ORSet[A, B] =
+    ORSet(this.values :+ ORSet.VersionedEntry[A, B](value, vectorClock))
+}
+
+object ORSet {
+  def empty[A, B] = ORSet[A, B](Vector.empty)
+
+  case class VersionedEntry[A, B](value: A, clock: VectorClock[B])
+}
--- a/src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
+++ b/src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
@@ -1,0 +1,57 @@
+package scalaz.ziokeeper.crdt
+
+import org.specs2.mutable.Specification
+import scalaz.ziokeeper.VectorClock
+
+class ORSetSpec extends Specification {
+  override def is =
+    s2"""
+    This specification is for ORSet
+
+    ORSet should
+      add element $e1
+      remove element $e2
+      not remove when element in not in $e3
+    """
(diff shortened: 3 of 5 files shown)


src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,64 @@
...
+  implicit class VectorClockOps[A](val v: VectorClock[A]) extends AnyVal {
+    def time(mid: A): Long = v.value.getOrElse(mid, 0L)
+
+    def increment(mid: A, value: Long): VectorClock[A] =
+      VectorClock(v.value.updated(mid, time(mid) + value))
+  }
+
+  sealed trait CompareVectorClock
+
+  case object TheSame extends CompareVectorClock
```
rabbitonweb: What's the difference between `TheSame` and `Concurrent`?

src/main/scala/scalaz/ziokeeper/crdt/ORSet.scala
```diff
@@ -0,0 +1,33 @@
+package scalaz.ziokeeper.crdt
+
+import scalaz.ziokeeper.VectorClock
+
+case class ORSet[A, B](values: Vector[ORSet.VersionedEntry[A, B]])
+
+object ORSet {
+  implicit class ORSetOps[A, B](val oRSet: ORSet[A, B]) extends AnyVal {
```
mijicd: Maybe it makes sense to have these as methods in `ORSet`.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,64 @@
...
+  def empty[A] = VectorClock[A](value = Map.empty)
+
+  implicit class VectorClockOps[A](val v: VectorClock[A]) extends AnyVal {
+    def time(mid: A): Long = v.value.getOrElse(mid, 0L)
+
+    def increment(mid: A, value: Long): VectorClock[A] =
+      VectorClock(v.value.updated(mid, time(mid) + value))
+  }
+
+  sealed trait CompareVectorClock
```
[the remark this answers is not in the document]
pshemass: I was trying to avoid ambiguity with scala.Ordering which is not the same and I need to implement Ordering anyway but I need to read one more paper to figure out what to do when I we have concurrent modifications of the same elements.

So do you have better name for that?

src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  case object TheSame extends CompareVectorClock
+
+  case object Before extends CompareVectorClock
+
+  case object After extends CompareVectorClock
+
+  case object Concurrent extends CompareVectorClock
+
+  def compare[A](left: VectorClock[A], right: VectorClock[A]): CompareVectorClock = {
```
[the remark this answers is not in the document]
pshemass: sure

src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.specs2.mutable.Specification
+import scalaz.ziokeeper.VectorClock
+
+class ORSetSpec extends Specification {
+  override def is =
+    s2"""
+    This specification is for ORSet
+
+    ORSet should
+      add and remove element $e1
```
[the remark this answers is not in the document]
pshemass: what do you mean?

src/main/scala/scalaz/ziokeeper/VectorClock.scala
pshemass: the idea behind vector clock is pretty simple 
- increment you local counter whatever you perform actions(e.g. add, remove, modify) as well as when you receive messages from other members of the cluster.

when we have cluster with three member with ids 111, 222, 333 vector clock initially will look like this on all nodes.
(111 -> 0, 222 -> 0, 333 -> 0) 

at this point the are all **TheSame**

imagine that 111 and 222 perform add action 
111 = (111 -> 1, 222 -> 0, 333 -> 0)
222 = (111 -> 0, 222 -> 1, 333 -> 0)
333 =  (111 -> 0, 222 -> 0, 333 -> 0)
111 and 222 are concurrent because you cannot say which happen first.

let me know if that clear enough.

src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+    val set     = ORSet.empty[Int, Int]
+    val value   = 100
+    //when
+    val withValue              = set.add(value, vc)
+    val (removedVcs, newORset) = withValue.localRemove(value)
+    val removed                = withValue.remove(value, removedVcs)
+    val notRemoved             = withValue.remove(200, removedVcs)
+    val notRemoved2            = withValue.remove(value, Vector(VectorClock.empty[Int].increment(localId, 2)))
+    //then
+    withValue.values must_=== Vector(ORSet.VersionedEntry(100, vc))
```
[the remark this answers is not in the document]
pshemass: how `containTheSameElementsAs` is different from this?

src/main/scala/scalaz/ziokeeper/crdt/ORSet.scala
pshemass: sure

src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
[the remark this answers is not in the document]
mijicd: Just wandering whether we could express this as a property-based test. Bad choice of wording from my side, I should have mentioned law :).

src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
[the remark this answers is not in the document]
mijicd: They're the same, I referred to this line wanting to point out we should stick to one style. The reason I suggested `containTheSameElementsAs` were these:

```scala
    removedVcs must haveSize(1)
    removedVcs.head must_=== vc
```

While all of these checks are in place, I'd rather avoid doing both checks (`must_==` is also fine). Also, `must haveSize(0)` can be replaced with `must be empty`.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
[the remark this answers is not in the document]
mijicd: I see. Not at the moment, let's leave it as it is, and change once we find the more suitable name.

src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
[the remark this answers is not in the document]
pshemass: you are right this is test is hard to follow. I broke it into couple of them. let me know what do you think.

src/test/scala/scalaz/ziokeeper/crdt/ORSetSpec.scala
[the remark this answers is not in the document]
pshemass: I was thinking about scalaCheck for that but i was too lazy...

rabbitonweb:
> Looks good to me! @jdegoes, @rabbitonweb any remarks that should be fixed before merging?

Give me 24h for my remarks

src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,63 @@
+package scalaz.ziokeeper
+
+import scala.annotation.tailrec
+import VectorClock._
+case class VectorClock[A](value: Map[A, Long]) extends AnyVal {
```
jdegoes: This is great for now, ultimately, `Map` won't be an efficient way to store or represent a vector clock. 👍

src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,63 @@
...
+  def time(mid: A): Long = value.getOrElse(mid, 0L)
+
+  def increment(mid: A, value: Long): VectorClock[A] =
```
jdegoes: This vector clock will only ever be incremented by `1`. We will maybe need to "set" certain components for other nodes.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,63 @@
...
+        case _ if leftBigger && rightBigger => (leftBigger, rightBigger)
+        case Nil                            => (leftBigger, rightBigger)
+        case h :: t =>
+          val diff = leftWithDefault(h) - rightWithDefault(h)
+          loop(t, leftBigger || diff > 0, rightBigger || diff < 0)
+      }
+
+    val keys = leftWithDefault.keySet.union(rightWithDefault.keySet)
+
+    loop(keys.toList, leftBigger = false, rightBigger = false) match {
```
jdegoes: This looks potentially non-deterministic (due to the use of `Set`), which will not satisfy laws for equality. 

Consider sorting by `A`, for which you'll need an ordering on `A`. This will be expensive and speaks to the fact that we ultimately need a different representation for `VectorClock`, namely, two arrays (one for `A`, sorted in order, and one for `Long`, with the same ordering). But for now the inefficiency can be ignored.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,63 @@
...
+      case (false, false) => TheSame
+      case (true, true)   => Concurrent
+      case (true, false)  => After
+      case (false, true)  => Before
+    }
+  }
+
+  def merge(right: VectorClock[A]): VectorClock[A] = {
```
jdegoes: Consider extracting out `mergeWith(right)((x: Long, y: Long) => Long)`, then you can implement this as `mergeWith(right)(_ max _)`.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
```diff
@@ -0,0 +1,63 @@
...
+            acc + (key -> math.max(leftWithDefault(key), rightWithDefault(key)))
+        }
+    )
+  }
+}
+
+object VectorClock {
+  def empty[A] = VectorClock[A](value = Map.empty)
+
+  sealed trait CompareVectorClock
```
jdegoes: This would more classically be called `Ordering` with `EQ`, `GT`, `LT`. I am not sure what `Concurrent` is.

src/main/scala/scalaz/ziokeeper/crdt/ORSet.scala
```diff
@@ -0,0 +1,31 @@
+package scalaz.ziokeeper.crdt
```
jdegoes: The new root package will probably be:

```scala
package dev.zio.keeper
```

Or maybe just:

```scala
package zio.keeper
```

src/main/scala/scalaz/ziokeeper/crdt/ORSet.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import scalaz.ziokeeper.VectorClock
+
+case class ORSet[A, B](values: Vector[ORSet.VersionedEntry[A, B]]) {
```
jdegoes: This is gonna be super slow. But I think that's OK. We should try to make it correct first, and we can optimize later. 👍

src/main/scala/scalaz/ziokeeper/crdt/ORSet.scala
mijicd: Shall we migrate the project to new organization?

jdegoes: Yes, no rush on that, but to avoid cluttering `scalaz` we should get all the active ZIO projects to the new org (and delete the remainder). 👍

mijicd: In that case, let's polish this PR, and I'll take care of the migration and packages cleanup.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
pshemass: you are right this is not most efficient implementation but I try to make this correct in the first place then optimize.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
pshemass: make sense

src/main/scala/scalaz/ziokeeper/VectorClock.scala
pshemass: but this is not exactly ordering in the classic `EQ`, `GT`, `LT` sense. 

this is explanation I gave to @rabbitonweb for a question about different between `TheSame` and `Concurrent` I hope that will help you understand `Concurrent`

> the idea behind vector clock is pretty simple

> increment you local counter whatever you perform actions(e.g. add, remove, modify) as well as when you receive and send messages from/to other members of the cluster.
> when we have cluster with three member with ids 111, 222, 333 vector clock initially will look like this on all nodes.
> (111 -> 0, 222 -> 0, 333 -> 0)
> 
> at this point the are all TheSame
> 
> imagine that 111 and 222 perform add action
> 111 = (111 -> 1, 222 -> 0, 333 -> 0)
> 222 = (111 -> 0, 222 -> 1, 333 -> 0)
> 333 = (111 -> 0, 222 -> 0, 333 -> 0)
> 111 and 222 are Concurrent because you cannot say which happen first.

I need to implement Ordering/PartialOrdering to be able to sort values in ORSet so I named this differently(pretty badly though) to avoid confusion.

The problem that I struggle now is how to make sure that result will be deterministic when we have `Concurrent` update on the same key.

src/main/scala/scalaz/ziokeeper/VectorClock.scala
pshemass: done

src/main/scala/scalaz/ziokeeper/VectorClock.scala
pshemass: renamed to `VectorClockOrdering`

src/main/scala/scalaz/ziokeeper/VectorClock.scala
pshemass: changed to `def increament(mid: A)`. we don't need set because merge should be enough.

PR #31, opened 2019-05-02
Trigger release on tags only

mijicd:


.circleci/config.yml
```diff
@@ -91,6 +91,6 @@ workflows:
             - test212_jdk11
           filters:
             branches:
-              only: master
+              ignore: /.*/
             tags:
               only: /^v[0-9]+(\.[0-9]+)*$/
```
ghostdogpr: See my comment here https://github.com/zio/zio-actors/pull/28#discussion_r280425328 that needs to be added for each step

mijicd: Done, please re-review.

PR #38, opened 2019-06-27
Draft new API

mijicd:
As an outcome of the conversation with @pshemass and @rabbitonweb, zio-keeper internals should be split into the following components:

- transport
- seeds discovery
- gossip
- public API

This PR will be used to sketch an API compatible with such separation. Once polished, subsequent issues for implementing each of the aforementioned components will be created.


src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  trait Transport {
+    def send[S, A](data: Message[S, A]): IO[Error, Unit]
+    def receive[S, A]: Stream[Error, Message[S, A]]
+  }
+
+  // TODO: maybe just an enumeration specifying the type of underlying data structure, e.g. CRDT
+  trait Metadata {}
+
+  trait Handle {
```
jdegoes: I'd remove this and Metadata in the initial implementation.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -0,0 +1,46 @@
...
+  final def join: ZIO[Credentials with Discovery with Transport with Metadata, Error, Handle] = ???
+}
+
+object Cluster {
+  // TODO: properly represent (see current 'develop' branch)
+  type Path[_, _]
+  type Type[_]
+
+  object Type {
+    def apply[A](implicit instance: Type[A]): Type[A] = instance
```
jdegoes: I'd remove this and Path in the initial implementation.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  trait Credentials {
+    // TODO: ways to obtain auth data
+  }
+
+  trait Discovery {
+    def discover: IO[Error, Set[InetAddress]]
```
jdegoes: The problem with this interface is that it does not permit notification on new members, etc. (in which case, `Managed[Error, Stream[Error, MembershipEvent]]` could work). Also possibly we should introduce the notion of "channel"—perhaps that's a "key" provided to "discover", i.e. give me all the nodes that identify with the specified key.

pshemass: this is interface that only provide list of seeds of cluster. If we want to use this in any production scenario we need to provide some way to discover them. typically it's not as easy as provide list of the ips. for instance when you are using kubernetes that will be call to kubernetes's dns or service for list of nodes.

jdegoes: OK, that makes sense. 👍

src/main/scala/zio/keeper/Cluster.scala
mijicd: Done.

src/main/scala/zio/keeper/Cluster.scala
mijicd: Done.

src/main/scala/zio/keeper/Cluster.scala
mijicd: As @pshemass explained, our rough idea is to handle `MembershipEvent` event stream internally.

PR #44, opened 2019-07-24
Transport layer

pshemass:
Example of usage

```
package zio.keeper

import zio.{ Chunk, IO }
import zio.keeper.Cluster.{ Credentials, Discovery }
import zio.keeper.Cluster.Transport.TCPTransport
import zio.nio.{ InetAddress, SocketAddress }
import zio.console._

object Node1 extends zio.App {

  def run(args: List[String]) =
    appLogic.fold(_ => 1, _ => 0)

  val config = new Credentials with TCPTransport with Discovery with zio.console.Console.Live {

    override def discover: IO[Error, Set[SocketAddress]] =
      InetAddress.localHost
        .flatMap(addr => SocketAddress.inetSocketAddress(addr, 5558))
        .map(Set(_: SocketAddress))
        .orDie
  }

  val appLogic = Cluster
    .join(5557)
    .provide(config)
    .flatMap(c => c.broadcast(Chunk.fromArray("foo".getBytes)).map(_ => c))
    .flatMap(
      c => c.receive.foreach(n => putStrLn(new String(n.payload.toArray)) *> c.send(n.payload, n.sender))
    )
}

object Node2 extends zio.App {

  def run(args: List[String]) =
    appLogic.fold(_ => 1, _ => 0)

  val config = new Credentials with TCPTransport with Discovery with zio.console.Console.Live {

    override def discover: IO[Error, Set[SocketAddress]] =
      InetAddress.localHost
        .flatMap(addr => SocketAddress.inetSocketAddress(addr, 5557))
        .map(Set(_: SocketAddress))
        .orDie
  }

  val appLogic = Cluster
    .join(5558)
    .provide(config)
    .flatMap(c => c.broadcast(Chunk.fromArray("bar".getBytes)).map(_ => c))
    .flatMap(
      c => c.receive.foreach(n => putStrLn(new String(n.payload.toArray)) *> c.send(n.payload, n.sender))
    )

}
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,6 +34,7 @@ lazy val zioKeeper = project
     libraryDependencies ++= Seq(
       "dev.zio"    %% "zio"                  % "1.0.0-RC11-1",
       "dev.zio"    %% "zio-streams"          % "1.0.0-RC11-1",
+      "dev.zio"    %% "zio-nio"              % "0.1.1",
       "org.specs2" %% "specs2-core"          % "4.7.0" % Test,
       "org.specs2" %% "specs2-scalacheck"    % "4.7.0" % Test,
       "org.specs2" %% "specs2-matcher-extra" % "4.7.0" % Test
--- a/src/main/scala/zio/keeper/Cluster.scala
+++ b/src/main/scala/zio/keeper/Cluster.scala
@@ -1,31 +1,279 @@
 package zio.keeper
 
-import java.net.InetAddress
+import java.math.BigInteger
+import java.util.UUID
 
-import zio.{ IO, ZIO }
+import zio._
+import zio.console.putStrLn
+import zio.keeper.Cluster.InternalProtocol.{ ProvideIdentity, RequestIdentity }
+import zio.keeper.Error.{ HandshakeError, NodeUnknown, SendError, SerializationError }
+import zio.nio._
+import zio.nio.channels.{ AsynchronousServerSocketChannel, AsynchronousSocketChannel }
 import zio.stream.Stream
-import zio.keeper.Cluster._
 
 trait Cluster {
-  final def join: ZIO[Credentials with Discovery with Transport, Error, Unit] = ???
+  def nodes: UIO[List[NodeId]]
+
+  def send(data: Chunk[Byte], receipt: NodeId): IO[Error, Unit]
+
+  def broadcast(data: Chunk[Byte]): IO[Error, Unit]
+
+  def receive: Stream[Error, Message]
 }
 
 object Cluster {
+
+  sealed trait InternalProtocol {
+    def serialize: IO[SerializationError, Chunk[Byte]]
+  }
+
+  object InternalProtocol {
+
+    case object RequestIdentity extends InternalProtocol {
+      override def serialize: IO[SerializationError, Chunk[Byte]] = ZIO.succeed(Chunk(1))
+    }
+
+    final case class ProvideIdentity(node: NodeId) extends InternalProtocol {
+      override def serialize: IO[SerializationError, Chunk[Byte]] =
+        ZIO.succeed(Chunk(2.toByte) ++ Chunk.fromArray(node.value.toString.getBytes))
+    }
+
+    final case class NotifyJoin(addr: InetAddress, port: Int) extends InternalProtocol {
+      override def serialize: IO[SerializationError, Chunk[Byte]] =
+        ZIO.succeed(
+          Chunk(3.toByte) ++
+            Chunk.fromArray(addr.address) ++
+            Chunk.fromArray(BigInteger.valueOf(port.toLong).toByteArray)
+        )
+    }
+
+    def deserialize(bytes: Chunk[Byte]): ZIO[Any, Throwable, InternalProtocol] =
+      if (bytes.isEmpty) {
+        ZIO.fail(SerializationError("Fail to deserialize message"))
+      } else {
+        val messageByte = bytes.drop(1)
+        bytes.apply(0) match {
+          case 1 =>
+            ZIO.succeed(RequestIdentity)
+          case 2 =>
+            ZIO.effect(ProvideIdentity(NodeId(UUID.fromString(new String(messageByte.toArray)))))
+          case 3 =>
+            for {
+              a   <- InetAddress.byAddress(messageByte.take(4).toArray)
+              res <- ZIO.effect(new BigInteger(messageByte.drop(4).toArray).intValue())
+            } yield NotifyJoin(a, res)
+        }
+      }
+
+  }
+
+  private def connectToSeed(publicAddr: InetAddress, port: Int, me: NodeId, seed: SocketAddress) =
+    for {
+      channel <- ZIO.accessM[Transport](_.connect(seed))
+      //initial handshake
+      bytes       <- RequestIdentity.serialize
+      ss          <- serializeMessage(me, bytes, 1)
+      _           <- channel.write(ss)
+      headerBytes <- channel.read(24)
+      byteBuffer  = new zio.nio.ByteBuffer(java.nio.ByteBuffer.wrap(headerBytes.toArray))
+      l1          <- byteBuffer.getLong
+      l2          <- byteBuffer.getLong
+      _           <- byteBuffer.getInt
+      payloadSize <- byteBuffer.getInt
+      payloadByte <- channel.read(payloadSize) // this should somehow be ignored when we cannot handle message because we are not able to handle protocol or version
+      _           = NodeId(new java.util.UUID(l1, l2))
+      m           <- InternalProtocol.deserialize(payloadByte)
+      res <- m match {
+              case InternalProtocol.ProvideIdentity(nodeId) =>
+                for {
+                  payload <- InternalProtocol.NotifyJoin(publicAddr, port).serialize
+                  bytes   <- serializeMessage(me, payload, 1)
+                  _       <- channel.write(bytes)
+                } yield (nodeId, channel)
+              case _ => ZIO.fail(HandshakeError("handshake error"))
+            }
+    } yield res
+
+  private def listenForClusterMessages(
+    currentNode: NodeId,
+    messageQueue: zio.Queue[Message],
+    initialNodes: Ref[Map[NodeId, AsynchronousSocketChannel]]
+  ) = {
+    for {
+      message <- messageQueue.take
+      payload <- InternalProtocol.deserialize(message.payload)
+      _ <- payload match {
+            case InternalProtocol.NotifyJoin(inetSocketAddress, port) =>
+              for {
+                client <- ZIO
+                           .accessM[Transport with zio.console.Console] { transport =>
+                             (SocketAddress.inetSocketAddress(inetSocketAddress, port) >>=
+                               transport.connect) <*
+                               putStrLn(
+                                 "connected with node [" + message.sender.value + "] " + inetSocketAddress.hostname + ":" + port
+                               )
+                           }
+                           .orDie
+                _ <- initialNodes.update(_.updated(message.sender, client)) //TODO here we should propagate membership event
+              } yield ()
+            case RequestIdentity =>
+              for {
+                payload <- ProvideIdentity(currentNode).serialize
+                bytes   <- serializeMessage(currentNode, payload, 1)
+                _       <- message.replyTo.write(bytes)
+              } yield ()
+            case _ => ZIO.unit
+          }
+    } yield ()
+  }.forever.fork
+
+  private def connectToCluster(localHost: InetAddress, port: Int, me: NodeId) =
+    for {
+      nodes <- zio.Ref.make(Map.empty[NodeId, AsynchronousSocketChannel])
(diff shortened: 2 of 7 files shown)


src/main/scala/zio/keeper/Message.scala
```diff
@@ -1,3 +1,86 @@
...
+case class Header(
+  protocolName: String,
+  protocolVersion: ProtocolDef.Version,
+  sender: NodeId,
+  messageType: Int,
+  payloadSize: Int
+)
+
+object Message {
+  val SIZE_OF_HEADER = 50
```
mijicd: Please rename to `HeaderSize`.

src/main/scala/zio/keeper/ProtocolDef.scala
```diff
@@ -0,0 +1,70 @@
...
+import zio.keeper.Error.ProtocolDef.{
+  CannotFindSerializerForMessage,
+  CannotFindSerializerForMessageId,
+  SerializationError
+}
+import zio.keeper.ProtocolDef.MessageDef
+
+import scala.reflect.ClassTag
+
+trait ProtocolDef[A] {
```
mijicd: I'd just rename it to `Protocol`, and `protocolName` to `name`.

src/main/scala/zio/keeper/ProtocolDef.scala
```diff
@@ -0,0 +1,70 @@
...
+  def protocolName: String // it would be cool to have byte limit for name
+
+  def version: ProtocolDef.Version
+
+  def schema: Schema[A]
+
+  protected def message[B <: A](
+    to: B => IO[SerializationError, Chunk[Byte]],
```
mijicd: Is it possible to "promote" `to` and `from` to trait-level methods and remove notion of `Schema`?

src/main/scala/zio/keeper/Message.scala
```diff
@@ -1,3 +1,86 @@
...
+
+import zio.keeper.Error.ProtocolDef.SerializationError
+import zio.nio.channels.AsynchronousSocketChannel
+import zio.{ Chunk, IO, ZIO }
+
+import scala.reflect.ClassTag
+
+final case class Message[A](header: Header, payload: A, replyTo: AsynchronousSocketChannel)
+
+case class Header(
```
mijicd: I'd keep `Header` in `Message`'s companion.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,355 @@
...
+  def send[B <: A](data: B, receipt: NodeId): IO[Error, Unit]
+
+  def broadcast[B <: A](data: B)(implicit classTag: ClassTag[B]): IO[Error, Unit]
+
+  def receive: Stream[Error, Message[A]]
 }
 
 object Cluster {
 
+  sealed trait InternalProtocol
```
mijicd: I'd separate it from the notion of protocol, to simplify naming. For example:

```scala
sealed trait Payload
object Payload {
  // TODO: define them
}

// TODO: see the remark below regarding serdes
val protocol = new Protocol[Event] { ... }
```
Visibility can be tuned, but I'd pick `private` wherever possible. The more I think about it, the more it looks as `Protocol` should be renamed to `Schema`.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,355 @@
...
+            (chunk: Chunk[Byte]) =>
+              (for {
+                a   <- InetAddress.byAddress(chunk.take(4).toArray)
+                res <- ZIO.effect(new BigInteger(chunk.drop(4).toArray).intValue())
+              } yield IamNewHere(a, res)).orDie //TODO proper error handling
+          )
+
+    }
+
+    def handshake(publicAddr: InetAddress, port: Int, me: NodeId, seed: SocketAddress) =
```
mijicd: Both `handshake` and `run` should probably be private methods on `Cluster` companion.

src/main/scala/zio/keeper/Message.scala
```diff
@@ -1,3 +1,86 @@
...
+  protocolVersion: ProtocolDef.Version,
+  sender: NodeId,
+  messageType: Int,
+  payloadSize: Int
+)
+
+object Message {
+  val SIZE_OF_HEADER = 50
+
+  def readHeader(chunk: Chunk[Byte]) = {
```
mijicd: I'd put `readHeader` and `writeHeader` to `Header`'s companion, and rename them to `deserialize` / `serialize` (or whatever notation we're using. I'd do the same for `Message`.

**edit:** `serialize` should live on instance itself not the companion.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,355 @@
...
+      for {
+        client <- ZIO.accessM[Transport] {
```
mijicd: Maybe rename `client` to `channel`, to be more explicit about its nature.

src/main/scala/zio/keeper/Message.scala
pshemass: done

src/main/scala/zio/keeper/Cluster.scala
pshemass: done

src/main/scala/zio/keeper/Cluster.scala
pshemass: i'm not sure about that but let's try.

build.sbt
```diff
@@ -27,15 +27,19 @@ inThisBuild(
...
     name := "zio-keeper",
     libraryDependencies ++= Seq(
-      "dev.zio"    %% "zio"                  % "1.0.0-RC9-4",
-      "dev.zio"    %% "zio-streams"          % "1.0.0-RC9-4",
-      "org.specs2" %% "specs2-core"          % "4.6.0" % Test,
-      "org.specs2" %% "specs2-scalacheck"    % "4.6.0" % Test,
-      "org.specs2" %% "specs2-matcher-extra" % "4.6.0" % Test
+      "dev.zio"         %% "zio"                  % "1.0.0-RC9-4",
+      "dev.zio"         %% "zio-streams"          % "1.0.0-RC9-4",
+      "com.nequissimus" %% "zio-slf4j"            % "0.2.0",
```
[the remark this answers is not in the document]
pshemass: done

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,355 @@
...
+        (msg: EchoMessage) => ZIO.succeed(Chunk.fromArray(msg.msg.getBytes)),
+        (chunk: Chunk[Byte]) => ZIO.succeed(EchoMessage(new String(chunk.toArray)))
+      )
+    )
+
+  }
+
+}
+
+object Node1 extends App {
```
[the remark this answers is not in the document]
pshemass: i will remove this for now and add this later

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,355 @@
...
+  object InternalProtocol {
+
+    case object PleaseIntroduceYourself extends InternalProtocol
+
+    case class Iam(node: NodeId) extends InternalProtocol
+
+    case class IamNewHere(addr: InetAddress, port: Int) extends InternalProtocol
+
+    val protocolDef = new ProtocolDef[InternalProtocol] {
+      override def protocolName: String = "int_protoc"
```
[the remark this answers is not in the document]
pshemass: sure

src/main/scala/zio/keeper/Cluster.scala
pshemass: done

src/main/scala/zio/keeper/Message.scala
pshemass: done. i really don't like adding methods to case classes(to be honest to any classes :))

src/main/scala/zio/keeper/Error.scala
```diff
@@ -1,7 +1,18 @@
 package zio.keeper
 
-sealed trait Error
+sealed trait Error extends Exception
 
 object Error {
+
+  object ProtocolDef {
```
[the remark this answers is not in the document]
pshemass: flattened

src/main/scala/zio/keeper/ProtocolDef.scala
pshemass: because we need to identify message type somehow it would be harder to put this in protocol.

mijicd:
@jdegoes any suggestions from your side? @pshemass implemented some of my suggestions, and we're not sure about the serialization part. It can be simpler, but it might not be suited for this PR. It would be great if you can provide some remarks / pointers that we could tackle.

build.sbt
```diff
@@ -27,13 +27,16 @@ inThisBuild(
 addCommandAlias("fmt", "all scalafmtSbt scalafmt test:scalafmt")
 addCommandAlias("check", "all scalafmtSbtCheck scalafmtCheck test:scalafmtCheck")
 
+addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
```
jdegoes: I would not recommend using this in ZIO projects. It will enable you to write code that does not work on Dotty. It's good for user-land apps, though.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,267 @@
...
+import zio.nio._
+import zio.nio.channels.{ AsynchronousServerSocketChannel, AsynchronousSocketChannel }
 import zio.stream.Stream
-import zio.keeper.Cluster._
 
-trait Cluster {
-  final def join: ZIO[Credentials with Discovery with Transport, Error, Unit] = ???
+import scala.reflect.ClassTag
+
+trait Cluster[A] {
```
jdegoes: I might suggest to remove `A` and just support messages of type `Chunk[Byte]`. It simplifies the design and de-conflates serialization/deserialization from the protocol concerns.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,267 @@
...
+  def nodes: UIO[List[NodeId]]
+
+  def send[B <: A](data: B, receipt: NodeId): IO[Error, Unit]
```
jdegoes: We should document the semantics of `send` under conditions of unreachability, downtime, etc. I think it makes sense to be conservative because a user can `.retry`, `.timeout`, etc., already using ZIO, so bundling timeouts and retry logic into `send` don't make sense unless there is a protocol-level reason for doing so.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,267 @@
...
+  def receive: Stream[Error, Message[A]]
 }
 
 object Cluster {
 
+  sealed trait InternalProtocol
+
+  object InternalProtocol {
+
+    case object PleaseIntroduceYourself extends InternalProtocol
```
jdegoes: `RequestIdentity`?

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,267 @@
...
+
+    final case class Iam(node: NodeId) extends InternalProtocol
```
jdegoes: `ProvideIdentity`?

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -1,31 +1,267 @@
...
+
+    final case class IamNewHere(addr: InetAddress, port: Int) extends InternalProtocol
```
jdegoes: `NotifyJoin`?

src/main/scala/zio/keeper/Cluster.scala
jdegoes: OTOH maybe this is more convenient from user-perspective, especially since the project will have minimal dependencies...

build.sbt
pshemass: sure I will remove it

src/main/scala/zio/keeper/Cluster.scala
pshemass: I agree let move this responsibility to library user.

src/main/scala/zio/keeper/Cluster.scala
pshemass: I agree, user will be responsible for handles errors and based on their logic retry or set timeout.

src/main/scala/zio/keeper/Cluster.scala
pshemass: names are hard :) I will change that to `RequestIdentity`

pshemass:
@mijicd @jdegoes I removed protocolDef and now we operate on `Chunk`. This move responsibility of serialization to the user.

mijicd:
@pshemass Could you remove build steps for 2.11? `zio-nio` doesn't support it at the moment, and we won't be able to provide builds for it until it's solved there.

PR #61, opened 2019-09-04
initial connection handshake improvements

pshemass:
this PR change the order of actions during initial connection to cluster. 

Most importantly node will not start listening on the the port if it's not in the cluster. 

Secondly we need to wait for confirmation of joining to the cluster from other nodes because we may broadcast message to nodes that don't have you on the list of nodes yet.

### DIFF
--- a/src/main/scala/zio/keeper/Cluster.scala
+++ b/src/main/scala/zio/keeper/Cluster.scala
@@ -2,11 +2,14 @@ package zio.keeper
 
 import java.math.BigInteger
 import java.util.UUID
+import java.util.concurrent.TimeUnit
 
 import zio._
+import zio.clock.Clock
 import zio.console.putStrLn
-import zio.keeper.Cluster.InternalProtocol.{ ProvideIdentity, RequestIdentity }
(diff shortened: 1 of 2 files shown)


src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -62,38 +69,49 @@ object Cluster {
               a   <- InetAddress.byAddress(messageByte.take(4).toArray)
               res <- ZIO.effect(new BigInteger(messageByte.drop(4).toArray).intValue())
             } yield NotifyJoin(a, res)
+          case 4 =>
```
mijicd: Let's replace the "magic" value with some constant.

src/main/scala/zio/keeper/Cluster.scala
```diff
@@ -47,6 +50,10 @@ object Cluster {
         )
     }
 
+    case object Ack extends InternalProtocol {
+      override def serialize: IO[SerializationError, Chunk[Byte]] = ZIO.succeed(Chunk(4))
```
mijicd: It can be val.

joshbrandy (commented):
Looking out for good work

PR #94, opened 2019-11-03
Upgrade zio nio version

mschuwalow:
This ended up doing a bit more than I would have liked, as I needed to touch quite a bit of implementation after everything became a managed. The gossip protocol changed into a very basic version of the SWIM protocol, but I'll open a seperate pr that implements it seperately.

### DIFF
--- a/src/main/scala/zio/keeper/Example.scala
+++ b/src/main/scala/zio/keeper/Example.scala
@@ -26,8 +26,8 @@ object Node1 extends zio.ManagedApp {
     .provide(config)
     .flatMap(
       c =>
-        //zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
-        c.broadcast(Chunk.fromArray("foo".getBytes)).map(_ => c).toManaged_
+        (zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
+        c.broadcast(Chunk.fromArray("foo".getBytes)).map(_ => c)).toManaged_
     )
     .flatMap(
       c =>
@@ -67,8 +67,8 @@ object Node2 extends zio.ManagedApp {
     .provide(config)
     .flatMap(
       c =>
-        //zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
-        c.broadcast(Chunk.fromArray("bar".getBytes)).as(c).toManaged_
+        (zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
+        c.broadcast(Chunk.fromArray("bar".getBytes)).as(c)).toManaged_
     )
     .flatMap(
       c =>
@@ -106,8 +106,8 @@ object Node3 extends zio.ManagedApp {
     .provide(config)
     .flatMap(
       c =>
-        //zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
-        c.broadcast(Chunk.fromArray("bar1".getBytes)).as(c).toManaged_
+        (zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
+        c.broadcast(Chunk.fromArray("bar1".getBytes)).as(c)).toManaged_
     )
     .flatMap(
       c =>
--- a/src/main/scala/zio/keeper/Example.scala
+++ b/src/main/scala/zio/keeper/Example.scala
@@ -27,16 +27,18 @@ object Node1 extends zio.ManagedApp {
     .flatMap(
       c =>
         (zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
-        c.broadcast(Chunk.fromArray("foo".getBytes)).map(_ => c)).toManaged_
+          c.broadcast(Chunk.fromArray("foo".getBytes)).map(_ => c)).toManaged_
     )
     .flatMap(
       c =>
-        c.receive.foreach(
-          n =>
-            putStrLn(new String(n.payload.toArray))
-              *> c.send(n.payload, n.sender)
-              *> zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS))
-        ).toManaged_
+        c.receive
+          .foreach(
+            n =>
+              putStrLn(new String(n.payload.toArray))
+                *> c.send(n.payload, n.sender)
+                *> zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS))
+          )
+          .toManaged_
     )
 
   def run(args: List[String]) =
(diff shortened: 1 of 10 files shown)


pshemass:
this version of zio-nio doesn't have fix that I proposed, right? 

could you please add some tests that prove that is working?

mschuwalow:
> this version of zio-nio doesn't have fix that I proposed, right?
> 
> could you please add some tests that prove that is working?

I'll double check the fix you mentioned and get back to you

mijicd:
@pshemass It doesn't as Maxim pushed a lot of it before the fix was merged, and it's not yet released as far as I'm aware of :).

src/main/scala/zio/keeper/Example.scala
```diff
@@ -18,37 +18,27 @@ object Node1 extends zio.App {
...
 
   val appLogic = Cluster
     .join(5557)
     .provide(config)
     .flatMap(
       c =>
-        //zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
-        c.broadcast(Chunk.fromArray("foo".getBytes)).map(_ => c)
+        (zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS)) *>
+          c.broadcast(Chunk.fromArray("foo".getBytes)).map(_ => c)).toManaged_
```
mijicd: I'd replace the last `map` with `as`.

mijicd:
Short update: `zio-nio:0.3.1` is published.

src/main/scala/zio/keeper/InternalCluster.scala
```diff
@@ -6,175 +6,292 @@ import zio.clock.Clock
...
-              } yield ()
-            case ProvideClusterState(state) =>
-              for {
-                currentClusterState <- gossipStateRef.get
-                diff                = currentClusterState.diff(state)
-                _                   = ZIO.foreach_(diff.local)(member => removeMember(member)) //this is incomplete - remote part of diif should be handle
-
-              } yield ()
-            case _ => putStrLn("unknown message: " + payload)
+      node <- nodesRef.get.map(_.get(to))
```
mijicd: This one's probably nicer as:

```scala
nodesRef.get.map(_.get(to)).flatMap {
  _.fold(ZIO.unit)(send => serializeMessage(localMember, payload, msgType) >>= send)
}
```

mschuwalow:
> this version of zio-nio doesn't have fix that I proposed, right?
> 
> could you please add some tests that prove that is working?

I believe the recent changes to zio-nio have introduced some regressions.
Please take a look here for details https://github.com/zio/zio-nio/issues/107

pshemass:
@mschuwalow this leak you reported can be fixed but Api won't change probably. I wouldn't believe that managed will safe the world always :)

@mijicd I still don't believe that this not working at all so without test we are not merging this :)

src/main/scala/zio/keeper/InternalCluster.scala
```diff
@@ -6,175 +6,292 @@ import zio.clock.Clock
...
-  nodesRef: Ref[Map[NodeId, AsynchronousSocketChannel]],
+  nodesRef: Ref[Map[NodeId, Chunk[Byte] => UIO[Unit]]],
   gossipStateRef: Ref[GossipState],
-  clusterMessageQueue: zio.Queue[Message],
-  userMessageQueue: zio.Queue[Message]
+  userMessageQueue: zio.Queue[Message],
+  subscribeToBroadcast: UIO[Stream[Nothing, Chunk[Byte]]],
+  publishToBroadCast: Chunk[Byte] => UIO[Unit],
+  msgOffset: Ref[Long],
+  acks: TMap[Long, Promise[Nothing, Unit]]
```
pshemass: is it working? :)

src/main/scala/zio/keeper/InternalCluster.scala
```diff
@@ -6,175 +6,292 @@ import zio.clock.Clock
...
+      }
+      loop.repeat(ZSchedule.spaced(30.seconds))
+    }
+
+  private def acceptConnectionRequests =
+    ZManaged.environment[Transport].flatMap(_.bind(localMember.addr).orDie).flatMap { server =>
+      ZIO
+        .uninterruptibleMask { restore =>
+          for {
+            res <- server.accept.reserve
```
pshemass: I see the trick now. I was trying to avoid this. I don't want to merge this if we have better solution. It's adding unnecessary complexity IMHO

src/main/scala/zio/keeper/Message.scala
```diff
@@ -1,6 +1,46 @@
...
+import zio.IO
+import zio.keeper.Error.DeserializationError
 
-final case class Message(sender: NodeId, payload: Chunk[Byte], replyTo: AsynchronousSocketChannel)
+final case class Message(
+  sender: NodeId,
+  payload: Chunk[Byte]
+)
+
+final case class OpenConnection(state: GossipState, address: Member) {
```
pshemass: I don't think that this belong this file

src/main/scala/zio/keeper/InternalCluster.scala
```diff
@@ -6,175 +6,292 @@ import zio.clock.Clock
...
+                  )
+                  .commit
+      _ <- promOpt.fold(ZIO.unit)(_.succeed(()).unit)
     } yield ()
 
-  private def accept =
+  private def runSwim =
+    Ref.make(0).flatMap { roundRobinOffset =>
+      val loop = gossipStateRef.get.map(_.members.filterNot(_ == localMember).toIndexedSeq).flatMap { nodes =>
+        if (nodes.size > 0) {
```
pshemass: ```node.nonEmpty```

mschuwalow:
> @mschuwalow this leak you reported can be fixed but Api won't change probably. I wouldn't believe that managed will safe the world always :)
> 
> @mijicd I still don't believe that this not working at all so without test we are not merging this :)

Sure, that's fair.

mschuwalow:
> @mschuwalow this leak you reported can be fixed but Api won't change probably. I wouldn't believe that managed will safe the world always :)
> 
> @mijicd I still don't believe that this not working at all so without test we are not merging this :)

Note that I believe that the updated AsynchronousServerSocketChannel#accept is inherently unsafe and cannot be fixed with this signature. I'll see what can be done though

pshemass:
@mschuwalow I was testing this locally and it looks like I cannot kill node with ctrl+c anymore :) check what happen when you kill one of the node too - infinite loop.

pshemass:
after discussion with @mschuwalow and @iravid and my own investigation I think we should merge and add separate tickets for:

1. graceful node shutdown with SIGTERM
2. better handle situation when one of the nodes die in cluster
3. Adding more tests for SWIM protocol

when @mschuwalow will merge changes in ```ZManage``` we will adjust our code.
let me know if you are in sync.

mijicd:
@pshemass agreed, I'll take care of merging / opening issues.
@mschuwalow we can start with the follow-ups today, if you agree.

mijicd:
@mschuwalow and I agreed to close this one and come with tightly-scoped smaller PRs. We'll create issues for each of these.

PR #96, opened 2019-11-10
factor out membership layer

mschuwalow:
as described in the https://github.com/zio/zio-keeper/milestone/1 it makes sense to have a seperate transport a membership / transport layer.

This pr is basically a high level draft of how these layers could look like and are quite inspired from the current ziokeeper code. I do have more stuff ready, but I would like us to start by discussing the high level api and design.

Transport layer should be rather self explanatory. I would highly like us to go for udp for the gossip messages though. The user (who will be internal for this stuff) should have control over what protocol he wants to use. The implementation I have in mind are essentially lazily created long lived tcp connection when necessary and udp otherwise. Listening should not distinguish between both I believe.

On the membership side I believe we should drop reliable broadcast as it is very hard to support / will have wonky semantics. Instead I propose that we allow the user to register messages that will spread as part of the underlying membership protocols gossip messages. Another thing that came up during experimenting with the current api is that I think it's very useful to support "wait for reply" at this level. I believe this can be very easily done by essentially adapting the address to be ```%nodeId-%optional converstation id-%maybe an index into the conversation?``` and handled transparently for the user.

I believe swim is a good fit for our membership protocol and also be a good basis for spreading of our crdt updates. I am thinking of having crdts be broadcast using registerBroadcast and implemented by having a clock for each broadcast (including membership protocol gossiping). Whenever a clock is ready to fire it checks for other clocks that are ready to fire and groups them into one message up to the udp limits. This design is taken from consuls memberlist and I think is very elegant solutions to this. 

There are a number of open questions here marked with todo that we should discuss. But all in all I think this forms a useful basis we can collaborate and split work on.

/cc @mijicd


membership/src/main/scala/zio/membership/NodeState.scala
```diff
@@ -0,0 +1,7 @@
+package zio.membership
+
+final case class NodeState(
+  node: NodeId,
+  incarnation: Int,
```
pshemass: what incarnation mean in this context?

membership/src/main/scala/zio/membership/Message.scala
```diff
@@ -0,0 +1,9 @@
+package zio.membership
+
+import zio.Chunk
+
+final case class Message(
```
pshemass: I think message should have ```correlationId``` that we can match all the message in conversation. WDYT?

membership/src/main/scala/zio/membership/transport/Transport.scala
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * TODO: Naming should be discussed.
+   * Our low level transport interface that allows sending messages either in fire-and-forget
+   * or in streaming style. Also allows listening to messages sends from other nodes.
+   *
+   * The default implementation of this should use both TCP and UDP and bind both protocols
+   * to the same port number on bind.
+   */
+  trait Service[R] {
+    def sendBestEffort(to: SocketAddress, msg: Chunk[Byte]): ZIO[R, Error, Unit]
```
pshemass: why do you think that we should distinct this in here? why not send send/bind only?

pshemass:
how do you know that is very hard to support broadcast?  i'm running 80 Hazelcast nodes cluster of production which pretty much provides reliable broadcast. I didn't have much problems with that for last 3 years... Apple App Store has probably 300 or more hazelcast nodes...

Other thing is do we want to replicate across entire cluster. why don't we split into partitions and then order updates in the partition? Having partition leader and backup that make thinks much simpler. 

Consul SWIM is very elegant solution but has a different usecase. They trying to detect if node went down in the cluster but as far I remember they use strong consistent kv value store to keep metadata in the cluster. The same as etcd in kubernetese. 

My point is if using the same mechanism for cluster failure detection and propagate data structure is good idea in the first place ?

membership/src/main/scala/zio/membership/transport/Transport.scala
mschuwalow: Performance mostly. I believe as much as possible should be done in udp for performance reasons, but for example user messages or large messages might be better send using tcp.

Also for example consul uses both in case a node is not reachable using udp but with tcp.

I'm not sure how exactly we want to do this, but I believe we should make both available.

membership/src/main/scala/zio/membership/Message.scala
mschuwalow: Do you mean for tracing? a combination of both should be usable for this. The idea is to have the node address + monotonically increasing counter so that every conversation has a unique id

membership/src/main/scala/zio/membership/NodeState.scala
mschuwalow: This should not be in there at this point. This is relevant for swim suspicion protocol and should move there if we decide to implement it

mschuwalow:
> how do you know that is very hard to support broadcast? i'm running 80 Hazelcast nodes cluster of production which pretty much provides reliable broadcast. I didn't have much problems with that for last 3 years... Apple App Store has probably 300 or more hazelcast nodes...

That is very cool 👍
I'm not familiar with hazelcast, but it does sound like a very impressive system.
Still I believe for broadcast you would need to rely on hazelcast infrastructure or similar. If we are going to build this from scratch we can only provide this at a much higher level in my opinion.
> 
> Other thing is do we want to replicate across entire cluster. why don't we split into partitions and then order updates in the partition? Having partition leader and backup that make thinks much simpler.

That is also an option. I believe we should wait for itamars requirements and our session this week either way.
> 
> Consul SWIM is very elegant solution but has a different usecase. They trying to detect if node went down in the cluster but as far I remember they use strong consistent kv value store to keep metadata in the cluster. The same as etcd in kubernetese.
> 
> My point is if using the same mechanism for cluster failure detection and propagate data structure is good idea in the first place ?

That is a good point. What makes their architecture attractive in my opinion is that they solve these issues in completely independently.
If I understand their architecture correctly on a high level they have their membership layer that basically has the same responsibilities as what we came up with for v.0.1. and have raft running on top of it to elect a leader for a replicated kv store. I don't believe that they have crdts in userland at all (they do have them as part of the membership protocol). 

I think this is quite nice for us as we also want to tackle these problems incrementally. We could reuse the membership layer for crdts, which I believe is not unusual for state-based (or the delta based crdts in the paper I linked, which target exactly this usecase) crdts. But we might also build this as a completely separate system on top of this, should that prove more viable.

Note that I have no strong attachment to this approach, it just seems like a sane way to get a lot of progress done while reusing a lot of ideas from a heavily battle-tested system.

membership/src/main/scala/zio/membership/Message.scala
pshemass: primarily I would like to have this for tracing. monotonically increased counter might be tricky if we will provide client library because we need to generate request id on client. this is important when someone need to debug what happen with his/her request.

membership/src/main/scala/zio/membership/transport/Transport.scala
pshemass: Good reasons. question is more about if this belongs to Transport. I thought that Transport is just ```UDP``` or ```TCP```. In your example with consul I would use both, not one combined.

I was thinking about something similar to this https://haskell-distributed.github.io/tutorials/1ch.html#sending-messages

`
import Network.Transport.TCP (createTransport, defaultTCPParameters)
....

main = do
  Right t <- createTransport "127.0.0.1" "10501" defaultTCPParameters
...
`

pshemass:
@mschuwalow consul's raft protocol is running on servers only. Servers has replication mechanism based on TCP. This is because of size of quorum (it's growing when you adding nodes) and that affect performance dramatically. They have gossip based only on client side.
https://www.consul.io/docs/internals/architecture.html

https://www.consul.io/docs/internals/consensus.html#raft-in-consul

But I'm glad that you are not very attached to this. Hopefully we will get some requirements from Itamar.

membership/src/main/scala/zio/membership/transport/Transport.scala
pshemass: we should only keep method ```send``` for now. we will add other later.

membership/src/main/scala/zio/membership/Membership.scala
```diff
@@ -0,0 +1,62 @@
...
+     */
+    def sendMessage(to: NodeId, data: Chunk[Byte]): ZIO[R, Error, Unit]
+
+    /**
+     * TODO: naming
+     * Send a message to a node. This will suspend until the message has been send.
+     * The returned message can be used to wait for the expected reply to this message and will
+     * fail once the timeout has passed.
+     */
+    def sendMessageReply(to: NodeId, data: Chunk[Byte], timeout: Duration): ZIO[R, Error, Promise[Unit, Message]]
```
pshemass: remove this for now

membership/src/main/scala/zio/membership/Membership.scala
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * The broadcasted messages will usually be piggybacked on the gossip messages of the membership
+     * protocol. If the gossip protocol would spread and a broadcasted message is ready for sending
+     * it will be embedded in the gossip message. Therefore choosing a multiple of the broadcast period
+     * of the gossip protocol leads to less network load here.
+     *
+     * Retransmits configures how often a new value should be propagated to cluster members. If a new value
+     * is registered by the user the old message will no longer be propagated.
+     */
+    def registerBroadcast(period: Duration, retransmits: Int): ZManaged[R, Error, Chunk[Byte] => UIO[Unit]]
```
pshemass: remove this for now

membership/src/main/scala/zio/membership/transport/Transport.scala
mschuwalow: One issue  I was thinking about is that some algorithms specifically need and reliable transport layer for failure detection.
This gets us to OOP land, but would do you think about having Transport <:< UnreliableTransport?

mschuwalow:
I've updated the pr. In it's current form it is again very close to do the original design in zio-keeper

mijicd:
@mschuwalow Let's merge them, as discussed over the meeting. Once this is in, I'll enable the 2.13 build as well. Other than that, looks great!

membership/src/main/scala/zio/membership/transport/Transport.scala
pshemass: your are optimizing for something that is not implemented or even planned. It would be better to just always keep MVP approach that we deliver minimal thing.

pshemass:
> @mschuwalow Let's merge them, as discussed over the meeting. Once this is in, I'll enable the 2.13 build as well. Other than that, looks great!

As we discussed membership should still be in core and we are going to decide what to do with this in the future.

mschuwalow:
Ok, one issue that I ran into just now is that core does not compile with current dependencies.
But I think it makes sense to keep it around until we reach feature parity

mijicd:
I'm merging this one in order to "unstuck" the other work. We can merge the modules and fix the deficiencies in subsequent PRs

PR #112, opened 2019-11-26
add tcp transport

### LINKED ISSUE #107: Upgrade ZIO to 1.0.0-RC17
Fix all errors present in the current codebase that prevent automatic migration.

mschuwalow:
closes #107

Initial version of tcp based transport

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,12 @@ inThisBuild(
         "Przemyslaw Wierzbicki",
         "rzbikson@gmail.com",
         url("https://github.com/pshemass")
+      ),
+      Developer(
+        "mschuwalow",
+        "Maxim Schuwalow",
+        "maxim.schuwalow@gmail.com",
+        url("https://github.com/mschuwalow")
       )
     ),
     pgpPassphrase := sys.env.get("PGP_PASSWORD").map(_.toArray),
@@ -58,13 +64,13 @@ lazy val membership = project
   .settings(
     name := "zio-membership",
     libraryDependencies ++= Seq(
-      "dev.zio"     %% "zio"              % "1.0.0-RC17",
-      "dev.zio"     %% "zio-streams"      % "1.0.0-RC17",
-      "dev.zio"     %% "zio-nio"          % "0.4.0",
-      "dev.zio"     %% "zio-macros-core"  % "0.6.0",
-      "com.lihaoyi" %% "upickle"          % "0.8.0",
-      "dev.zio"     %% "zio-test"         % "1.0.0-RC17" % "test",
-      "dev.zio"     %% "zio-test-sbt"     % "1.0.0-RC17" % "test"
+      "dev.zio"     %% "zio"             % "1.0.0-RC17",
+      "dev.zio"     %% "zio-streams"     % "1.0.0-RC17",
+      "dev.zio"     %% "zio-nio"         % "0.4.0",
+      "dev.zio"     %% "zio-macros-core" % "0.6.0",
+      "com.lihaoyi" %% "upickle"         % "0.8.0",
+      "dev.zio"     %% "zio-test"        % "1.0.0-RC17" % "test",
+      "dev.zio"     %% "zio-test-sbt"    % "1.0.0-RC17" % "test"
     ),
     testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework"))
   )
--- a/build.sbt
+++ b/build.sbt
@@ -58,12 +58,13 @@ lazy val membership = project
   .settings(
     name := "zio-membership",
     libraryDependencies ++= Seq(
-      "dev.zio"     %% "zio"          % "1.0.0-RC16",
-      "dev.zio"     %% "zio-streams"  % "1.0.0-RC16",
-      "dev.zio"     %% "zio-nio"      % "0.3.0",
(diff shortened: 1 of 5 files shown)


membership/src/main/scala/zio/membership/Error.scala
```diff
@@ -1,9 +1,24 @@
...
+import zio.duration.Duration
+import zio.nio.SocketAddress
 
 // TODO: define error hierarchy
-sealed trait Error extends Exception
+sealed abstract class Error(msg: String = "", cause: Throwable = null) extends Exception(msg, cause)
 
-sealed trait SendError    extends Error
-sealed trait ReceiveError extends Error
+sealed abstract class SendError(msg: String = "", cause: Throwable = null) extends Error(msg, cause)
```
pshemass: We could simplify this to ```TransportError```

```
sealed abstract class TransportError(msg: String = "", cause: Throwable = null) extends Error(msg, cause)

final case class ExceptionThrown(exc: Throwable) extends TransportError(cause = exc)
final case class RequestTimeout(timeout: Duration) extends TransportError(msg = s"Request timeout $timeout.")
final case class ConnectionTimeout(timeout: Duration) extends TransportError(msg = s"Connection timeout $timeout.")
final case class BindFailed(addr: SocketAddress, exc: Throwable)
       extends TransportError(msg = s"Failed binding to address $addr.", cause = exc)
 
```

membership/src/main/scala/zio/membership/transport/tcp.scala
```diff
@@ -0,0 +1,82 @@
...
+                                  } yield data
+                                )
+                                .timeout(connectionTimeout)
+                                .catchAllCause(_ => ZStream.empty)
+                            }
+                          }
+                        }
+                        .mapError(ReceiveError.ExceptionThrown)
+                      // hack to properly handle interrupts
+                      messages.merge(ZStream.never.ensuring(close))
```
pshemass: can we add test for interrupts?

build.sbt
```diff
@@ -58,12 +64,13 @@ lazy val membership = project
...
-      "dev.zio"     %% "zio-streams"  % "1.0.0-RC16",
-      "dev.zio"     %% "zio-nio"      % "0.3.0",
-      "com.lihaoyi" %% "upickle"      % "0.8.0",
-      "dev.zio"     %% "zio-test"     % "1.0.0-RC16" % "test",
-      "dev.zio"     %% "zio-test-sbt" % "1.0.0-RC16" % "test"
+      "dev.zio"     %% "zio"             % "1.0.0-RC17",
+      "dev.zio"     %% "zio-streams"     % "1.0.0-RC17",
+      "dev.zio"     %% "zio-nio"         % "0.4.0",
+      "dev.zio"     %% "zio-macros-core" % "0.6.0",
+      "com.lihaoyi" %% "upickle"         % "0.8.0",
```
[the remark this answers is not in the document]
mijicd: Chosen as the most stable among the existing ones. There were a couple of other candidates, but we put them aside for various reasons (at least until we come up with our own codec library :)).

membership/src/test/scala/zio/membership/transport/TcpTransportSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+
+            environment >>> Live.live(for {
+              port   <- freePort
+              addr   <- SocketAddress.inetSocketAddress(port)
+              chunk  <- readOne(addr).fork
+              _      <- send(addr, payload).delay(connectDelay)
+              result <- chunk.join
+            } yield assert(result, equalTo(payload)))
+          }
+        } @@ flaky @@ timeout(120.seconds) // this one is slow
```
mijicd: We should probably follow ZIO's approach and allocate machine with more resources for tests like this one. Also, we can consider having an alternate test strategy for long-running tests (i.e. not on every PR).

However, that should be done in the follow-up.

membership/src/test/scala/zio/membership/transport/TcpTransportSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+import zio.clock.Clock
+import zio.duration._
+import zio.macros.delegate._
+import zio.nio.SocketAddress
+import zio.test.environment.TestClock
+import zio.test.environment.Live
+import zio.test.TestAspect._
+
+object TransportSpec
+    extends DefaultRunnableSpec({
```
mijicd: I'm looking forward to having inheritance-based specs here.

membership/src/test/scala/zio/membership/transport/TcpTransportSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+      // todo: actually find free port
+      def freePort = ZIO.succeed(8010)
```
mijicd: It can be val.

mijicd:
closes #107

membership/src/test/scala/zio/membership/transport/TcpTransportSpec.scala
mschuwalow: Yeah, they are much less awkward

build.sbt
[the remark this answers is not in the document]
pshemass: that's great! writing those serializers are unnecessary burden at this stage so I agree that we should use something for that.

PR #118, opened 2019-11-27
Feature/hyparview

mschuwalow:
This is super wip, but might be already interesting to discuss things.


build.sbt
```diff
@@ -72,5 +72,6 @@ lazy val membership = project
       "dev.zio"     %% "zio-test"        % "1.0.0-RC17" % "test",
       "dev.zio"     %% "zio-test-sbt"    % "1.0.0-RC17" % "test"
     ),
-    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework"))
+    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework")),
+    addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
```
pshemass: John insisted to not use this

mschuwalow: Interesting. Do you happen to know why?

mschuwalow:
There is still quite a lot that I want to do here (tests, better error handling, ...), but this
is now a full implementation of hyparview and seems to work correctly from manual testing.

I believe that at this point no major changes should be necessary, so I think we can begin reviewing this.

build.sbt
pshemass: Yeah I know it was the first thing I added to the project when we started :) @jdegoes said back then that this will be problematic with dotty and it's good for client code but not for libs.

pshemass:
> There is still quite a lot that I want to do here (tests, better error handling, ...), but this
> is now a full implementation of hyparview and seems to work correctly from manual testing.
> 
> I believe that at this point no major changes should be necessary, so I think we can begin reviewing this.

it looks good but we need thorough tests because SWIM was very buggy :) I will move some of the stuff from your PR related to serialization to the core soon.

build.sbt
mschuwalow: alright, I'll remove it. It's not used much here anyway

mschuwalow:
> > There is still quite a lot that I want to do here (tests, better error handling, ...), but this
> > is now a full implementation of hyparview and seems to work correctly from manual testing.
> > I believe that at this point no major changes should be necessary, so I think we can begin reviewing this.
> 
> it looks good but we need thorough tests because SWIM was very buggy :) I will move some of the stuff from your PR related to serialization to the core soon.

agreed 👍

build.sbt
mijicd: Also on the line of being explicit rather than implicit inside lib code.

build.sbt
```diff
@@ -64,15 +64,18 @@ lazy val membership = project
...
-      "dev.zio"                %% "zio-test-sbt"            % "1.0.0-RC17" % Test
+      "dev.zio"     %% "zio"             % "1.0.0-RC17",
+      "dev.zio"     %% "zio-streams"     % "1.0.0-RC17",
+      "dev.zio"     %% "zio-nio"         % "0.4.0",
+      "dev.zio"     %% "zio-macros-core" % "0.6.0",
+      "com.lihaoyi" %% "upickle"         % "0.8.0",
+      "dev.zio"     %% "zio-test"        % "1.0.0-RC17" % Test,
+      "dev.zio"     %% "zio-test-sbt"    % "1.0.0-RC17" % Test,
+      ("com.github.ghik" % "silencer-lib" % "1.4.4" % Provided).cross(CrossVersion.full),
+      compilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1"),
```
mijicd: Nuke as discussed :).

build.sbt
```diff
@@ -64,15 +64,18 @@ lazy val membership = project
...
+      compilerPlugin(("com.github.ghik" % "silencer-plugin" % "1.4.4").cross(CrossVersion.full))
```
mijicd: This can go to `BuildHelper` as it's done in other projects in the organization. The same applies to `silencer-lib`.

membership/src/main/scala/zio/membership/Main.scala
```diff
@@ -0,0 +1,44 @@
...
+          i => (i.hostString, i.port),
+          { case (host: String, port: Int) => unsafeRun(SocketAddress.inetSocketAddress(host, port)) }
+        )
+
+    SocketAddress
+      .inetSocketAddress(8000)
+      .flatMap { addr =>
+        (env >>> HyParView(
+          addr,
+          10,
```
mijicd: Let's try to reduce magic values.

build.sbt
```diff
@@ -64,15 +64,18 @@ lazy val membership = project
...
-      "dev.zio"                %% "zio-nio"                 % "0.4.0",
-      "dev.zio"                %% "zio-macros-core"         % "0.6.0",
-      "com.lihaoyi"            %% "upickle"                 % "0.8.0",
-      "org.scala-lang.modules" %% "scala-collection-compat" % "2.1.2",
-      "dev.zio"                %% "zio-test"                % "1.0.0-RC17" % Test,
-      "dev.zio"                %% "zio-test-sbt"            % "1.0.0-RC17" % Test
+      "dev.zio"     %% "zio"             % "1.0.0-RC17",
+      "dev.zio"     %% "zio-streams"     % "1.0.0-RC17",
+      "dev.zio"     %% "zio-nio"         % "0.4.0",
+      "dev.zio"     %% "zio-macros-core" % "0.6.0",
```
mijicd: I believe we're good at these for now. However, they might be tricky in the future, and I'm not sure whether the benefit accommodates the cost.

membership/src/main/scala/zio/membership/Membership.scala
```diff
@@ -7,34 +7,41 @@ import zio.stream._
  * The main entrypoint to the membership library. Allows querying the cluster state
  * and sending messages to members.
  */
-trait Membership {
-  val membership: Membership.Service[Any]
+trait Membership[T] {
```
mijicd: Nitpick: A instead of T :).

membership/src/main/scala/zio/membership/hyparview/Env.scala
```diff
@@ -0,0 +1,126 @@
...
+      passiveView.size.map(_ >= cfg.passiveViewCapacity)
+
+    lazy val dropOneActiveToPassive =
+      for {
+        active  <- activeView.keys
+        dropped <- selectOne(active)
+        _       <- STM.foreach(dropped)(addNodeToPassiveView(_))
+      } yield dropped
+
+    def addNodeToActiveView(node: T, connection: ChunkConnection) =
```
mijicd: Please add the return type to these.

membership/src/main/scala/zio/membership/hyparview/Env.scala
```diff
@@ -0,0 +1,126 @@
...
+      if (node == myself) STM.succeed(None)
+      else {
+        for {
+          contained <- activeView.contains(node)
+          dropped <- if (contained) STM.succeed(None)
+                    else {
+                      for {
```
mijicd: Please extract to helper.

membership/src/main/scala/zio/membership/hyparview/Env.scala
```diff
@@ -0,0 +1,126 @@
...
+          index    <- pickRandom(values.size)
+          selected = values(index)
+        } yield Some(selected)
+      }
+
+    def selectN[A](values: List[A], n: Int) =
+      if (values.isEmpty) STM.succeed(Nil)
+      else {
+        def go(remaining: List[A], toPick: Int, acc: List[A]): STM[Nothing, List[A]] =
+          (remaining, toPick) match {
```
mijicd:
```scala
if (remaining.isEmpty || toPick == 0) STM.succeed(acc) 
else { ... }
```

Slightly cleaner imho.

membership/src/main/scala/zio/membership/hyparview/HyParView.scala
```diff
@@ -0,0 +1,97 @@
...
+import com.github.ghik.silencer.silent
+import zio.console.Console
+import zio.membership.Membership
+import zio.membership.ByteCodec
+import zio.duration._
+import zio.clock.Clock
+
+object HyParView {
+
+  @silent("deprecated")
```
mijicd: What's deprecated?

mijicd:
Also, as an addition to my previous comment, we should cut down on for-comprehensions. They become a mess when they're intertwined with nested comprehensions, and branches, plus the obvious deficiencies they anyhow have.

pshemass:
> Also, as an addition to my previous comment, we should cut down on for-comprehensions. They become a mess when they're intertwined with nested comprehensions, and branches, plus the obvious deficiencies they anyhow have.

I don't know if I agree with that. With for comprehension it's so easy to follow code like `do` in Haskell. I agree that nested for-comp should not be allowed.

mijicd:
I was mostly aiming at these, comprehensions inside comprehensions, mixed with branchings, plus the ones that are handled by better-monadic-for. In these situations, we should either remove them or extract them to helpers.

membership/src/main/scala/zio/membership/hyparview/HyParView.scala
mschuwalow: Stream in 2.13. I've separated that stuff out

membership/src/main/scala/zio/membership/Membership.scala
mschuwalow: probably makes sense. I've tried to be consistent with T representing the node identifier, but probably it does not buy much

membership/src/main/scala/zio/membership/hyparview/Env.scala
mschuwalow: will do

membership/src/main/scala/zio/membership/hyparview/Env.scala
mschuwalow: will do

build.sbt
mschuwalow: What do you mean?

membership/src/main/scala/zio/membership/Main.scala
mschuwalow: This is just an example main in order to play with cluster behavior. I'll remove them completely soonish

build.sbt
mschuwalow: Will do this in a separate pr.

mschuwalow:
Let's figure out what the big remaining issues are and then lets merge this already if possible. I can then address the remaining issues in incremental prs and we don't have the risk of this diverging too far from master

membership/src/main/scala/zio/membership/Main.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  override def run(args: List[String]) = {
+    val env =
+      ZManaged.environment[ZEnv] @@
+        tcp.withTcpTransport(64, 100.seconds, 100.seconds)
+    implicit val rw: ReadWriter[InetSocketAddress] =
+      macroRW[(String, Int)]
+        .bimap[InetSocketAddress](
+          i => (i.hostString, i.port),
+          { case (host: String, port: Int) => unsafeRun(SocketAddress.inetSocketAddress(host, port)) }
```
pshemass: I don't like that :(

mschuwalow: Yeah, this will disappear. The mains are just a basic example, I'll need to take a look how to properly serialize this stuff

PR #122, opened 2019-11-30
Enable 2.13 build steps

mijicd:


build.sbt
```diff
@@ -48,13 +48,14 @@ lazy val keeper = project
   .settings(stdSettings("zio-keeper"))
   .settings(
     libraryDependencies ++= Seq(
-      "dev.zio"     %% "zio"             % "1.0.0-RC17",
```
mschuwalow: We should start introducing a constant for zio versions. :D

membership/src/test/scala/zio/membership/transport/TcpTransportSpec.scala
```diff
@@ -33,8 +30,7 @@ object TransportSpec
               bind(addr).take(1).runCollect.map(_.head)
 
             environment >>> Live.live(for {
-              port   <- freePort
-              addr   <- SocketAddress.inetSocketAddress(port)
+              addr   <- SocketAddress.inetSocketAddress(0)
```
mschuwalow: Nice fix!

build.sbt
pshemass: I added this to https://github.com/zio/zio-keeper/issues/115

membership/src/test/scala/zio/membership/transport/TcpTransportSpec.scala
mijicd: Still fiddle around it, should be done tonight.

PR #125, opened 2019-12-04
refactor before membership events

pshemass:
I had to clean up couple of the things before introducing membership events


examples/src/main/scala/zio/keeper/example/Example.scala
```diff
@@ -51,7 +46,7 @@ object Node1 extends zio.ManagedApp {
     env.toManaged_
       .flatMap(e => appLogic.provide(e))
       .fold(ex => {
-        println(ex)
+        println("eeee: " + ex.msg)
```
mijicd: We should remove this eventually :)

keeper/src/main/scala/zio/keeper/discovery/K8DnsDiscovery.scala
```diff
@@ -17,18 +17,19 @@ import zio.{ IO, ZIO, keeper }
  * Headless service is a service of type ClusterIP with the clusterIP property set to None.
  *
  */
-trait K8DnsDiscovery extends Discovery {
+abstract class K8DnsDiscovery(console: Console) extends Discovery.Service[Any] {
```
mijicd: What's the reason for injecting it this way?

pshemass: to be honest I didn't know how to do this other way. If you can give me a hand here it would be great. sometimes I'm lost with those environments tricks :)

examples/src/main/scala/zio/keeper/example/Example.scala
pshemass: done :)

examples/src/main/scala/zio/keeper/example/Example.scala
```diff
@@ -5,152 +5,74 @@ import java.util.concurrent.TimeUnit
...
+              n =>
+                putStrLn(new String(n.payload.toArray))
+                  *> c.send(n.payload, n.sender).ignore
+                  *> zio.ZIO.sleep(zio.duration.Duration(5, TimeUnit.SECONDS))
+            )
+            .toManaged_
+      )
       .fold(ex => {
-        println(ex)
+        println(s"exit with error: $ex")
```
mijicd: This one should also go away, but not now.

PR #131, opened 2019-12-10
Fix gossip problem

pshemass:


pshemass:
> Looks good. I would prefer not to spend to much time on this as it will be replaced soonish

who told you that it gonna be replaced? We agreed that we will merge your stuff eventually but it doesn't mean that we we gonna remove this one. I want to have more than one membership protocol. beside that we need to show something on demo in London :)

PR #137, opened 2019-12-17
Order the functions alphabetically , and extract the versions of zio

naseem91:
I'm trying to clean the code. I plan to :
- ensure return type for each function 
- Change def to val if applicable 
- Order the functions alphabetically
- Extract the versions of ZIO

What we have done : 
- Order the functions alphabetically
- Extract the versions of ZIO

ensure return type for each function 
and change def to val if applicable 
require more understanding so I'm looking for more documentation .


.gitignore
```diff
@@ -1,5 +1,6 @@
 target
 .sbtopts
 project/.sbt
+.idea
```
mijicd: Please read the comment below the line you just added.

keeper/src/main/scala/zio/keeper/membership/Member.scala
```diff
@@ -12,12 +12,12 @@ final case class NodeId(value: UUID) extends AnyVal
...
   implicit val ordering: Ordering[NodeId] = Ordering.by(_.value)
 
-  def generateNew: NodeId =
-    NodeId(UUID.randomUUID())
+  val generateNew: NodeId = NodeId(UUID.randomUUID())
 }
 
 final case class Member(nodeId: NodeId, addr: NodeAddress) {
-  override def toString: String = s"nodeId: ${nodeId.value}, ip: ${addr.ip.mkString(".")}, port: ${addr.port}"
+  //conv
```
mijicd: Please remove the comment. Also, I'd leave `toString` as def.

keeper/src/main/scala/zio/keeper/membership/Membership.scala
```diff
@@ -13,13 +13,17 @@ object Membership {
   trait Service[R] {
     def broadcast(data: Chunk[Byte]): ZIO[R, Error, Unit]
 
-    def events: ZStream[R, Error, MembershipEvent]
+    //conv
```
mijicd: Please remove all `conv` comments.

keeper/src/main/scala/zio/keeper/transport/transport.scala
```diff
@@ -21,14 +21,19 @@ object Transport {
 }
 
 sealed trait Channel {
-  def isOpen: ZIO[Any, TransportError, Boolean]
 
+  //conv
+  val isOpen: ZIO[Any, TransportError, Boolean]
+
+  //conv
```
mijicd: `close` can be commented out as well.

.gitignore
mijicd: Also, please check out the diff. This didn't really ignore anything. You'll have to add the trailing slash for this to be valid, but please not here.

.gitignore
```diff
@@ -1,6 +1,7 @@
 target
 .sbtopts
 project/.sbt
-.idea
+.idea/
```
mijicd: One more time, **please** read the link below this line.

mijicd:
@naseem91 It's getting there, but there are things that are repetitive:

- Please read the link provided in `.gitignore`. It provides an explanation how to properly exclude IDE artifacts globally on your machine.
- Please format the sources. Note that all CI steps have to pass before considering to merge.

mijicd:
Closing the PR as per an agreement with @naseem91. Instead of having one large PR tackling multiple concerns, we should strive for having multiple smaller ones that are easier to review.

PR #144, opened 2020-01-02
Extracting zio versions to a variable in build.sbt  

naseem91:


mijicd:
This PR doesn't only affect ZIO version. It contains most of the changes introduced by the closed PR (#137) like alphabetizing methods, and it doesn't ignore the editor files, nor does it passes the formatting step.

PR #160, opened 2020-01-27
Added correlaction_id to the message.

klangner:


keeper/src/main/scala/zio/keeper/Message.scala
```diff
@@ -7,34 +7,46 @@ import zio.nio.Buffer
 import zio.{ Chunk, IO }
 
 final case class Message(
+  id: String,
```
pshemass: I want this to be ```UUID```

keeper/src/main/scala/zio/keeper/membership/package.scala
```diff
@@ -20,6 +20,6 @@ package object membership extends Membership.Service[Membership] {
   override def receive: ZStream[Membership, Error, Message] =
     ZStream.unwrap(ZIO.access[Membership](_.membership.receive))
 
-  override def send(data: Chunk[Byte], receipt: NodeId): ZIO[Membership, Error, Unit] =
-    ZIO.accessM[Membership](_.membership.send(data, receipt))
+  override def send(messageId: String, data: Chunk[Byte], receipt: NodeId): ZIO[Membership, Error, Unit] =
```
pshemass: This is not good idea. this need to be like correlationId in messaging systems. relying on users to make them unique would be silly.

keeper/src/main/scala/zio/keeper/membership/Membership.scala
```diff
@@ -21,7 +21,7 @@ object Membership {
 
     def receive: ZStream[R, Error, Message]
 
-    def send(data: Chunk[Byte], receipt: NodeId): ZIO[R, Error, Unit]
+    def send(messageId: String, data: Chunk[Byte], receipt: NodeId): ZIO[R, Error, Unit]
```
pshemass: please remove this

keeper/src/main/scala/zio/keeper/membership/package.scala
```diff
@@ -5,8 +5,8 @@ import zio.{ Chunk, ZIO }
 
 package object membership extends Membership.Service[Membership] {
 
-  override def broadcast(data: Chunk[Byte]): ZIO[Membership, Error, Unit] =
-    ZIO.accessM[Membership](_.membership.broadcast(data))
+  override def broadcast(messageId: String, data: Chunk[Byte]): ZIO[Membership, Error, Unit] =
```
pshemass: remove this

keeper/src/main/scala/zio/keeper/membership/Membership.scala
```diff
@@ -11,7 +11,7 @@ trait Membership {
 object Membership {
 
   trait Service[R] {
-    def broadcast(data: Chunk[Byte]): ZIO[R, Error, Unit]
+    def broadcast(messageId: String, data: Chunk[Byte]): ZIO[R, Error, Unit]
```
pshemass: remove this

keeper/src/main/scala/zio/keeper/membership/SWIM.scala
```diff
@@ -39,8 +39,8 @@ final class SWIM(
   override val localMember: ZIO[Any, Nothing, Member] =
     ZIO.succeed(localMember_)
 
-  override def broadcast(data: Chunk[Byte]): IO[Error, Unit] =
-    serializeMessage(localMember_, data, 2).flatMap[Any, Error, Unit](publishToBroadcast).unit
+  override def broadcast(messageId: String, data: Chunk[Byte]): IO[Error, Unit] =
```
pshemass: remove this

keeper/src/main/scala/zio/keeper/membership/SWIM.scala
```diff
@@ -49,15 +49,15 @@ final class SWIM(
   override def receive: Stream[Error, Message] =
     zio.stream.Stream.fromQueue(userMessageQueue)
 
-  override def send(data: Chunk[Byte], receipt: NodeId): IO[Error, Unit] =
-    sendMessage(receipt, 2, data)
+  override def send(messageId: String, data: Chunk[Byte], receipt: NodeId): IO[Error, Unit] =
```
pshemass: remove this from API

keeper/src/main/scala/zio/keeper/membership/SWIM.scala
```diff
@@ -49,15 +49,15 @@ final class SWIM(
...
+    sendMessage(receipt, 2, messageId, data)
 
-  private def sendMessage(to: NodeId, msgType: Int, payload: Chunk[Byte]) =
+  private def sendMessage(to: NodeId, msgType: Int, messageId: String, payload: Chunk[Byte]) =
     for {
       node <- nodeChannels.get.map(_.get(to))
       _ <- node match {
             case Some(channel) =>
-              serializeMessage(localMember_, payload, msgType) >>= channel.send
+              serializeMessage(messageId, localMember_, payload, msgType) >>= channel.send
```
pshemass: you we can generate ```UUID``` here.

keeper/src/main/scala/zio/keeper/membership/SWIM.scala
```diff
@@ -327,7 +327,7 @@ final class SWIM(
     for {
       _       <- logger.info(s"sending $msg")
       payload <- msg.serialize
-      msg     <- serializeMessage(localMember_, payload, 1)
+      msg     <- serializeMessage("", localMember_, payload, 1)
```
pshemass: we should generate UUID here it's important to have tracing for internal messages.

keeper/src/main/scala/zio/keeper/membership/package.scala
klangner: I probably misunderstood how it should work. 
I thought that we want to be able to track the same message send by different servers. If every time we send message we will assign it a new correlation_id, we won't be able to track the message across many hops.
But if this is for internal tracking for single send, then ok, I will remove it from the API.

pshemass: we should have something in API like reply to message. then we should keep track of conversationID but we don't have it yet. 

you should look into SWIM and make sure that when we reply for ping and other internal messages we should pass this id along.

klangner: Ok. I did the changes. Is it ok now? Or should I change the API of all function which send internal message and add this correlactionId there??

pshemass: when you look into method `handleClusterMessages` in `zio.keeper.membership.SWIM` you will see that we are replying for `Ping` and `PingReq` that are part of `SWIM` protocol. Ideally we will use the same `correlationId` for response. Could you add that this?

klangner: ok

PR #165, opened 2020-02-03
add ByteCodec laws

mschuwalow:


membership/src/main/scala/zio/membership/hyparview/TimeToLive.scala
```diff
@@ -2,14 +2,14 @@ package zio.membership.hyparview
 
 import upickle.default._
 
-final private[hyparview] case class TimeToLive(count: Int) {
+final case class TimeToLive(count: Int) {
```
mijicd: This can be a value class. Also, technically speaking, it shouldn't be negative.

Issue #133, opened 2019-12-13
Add UDP transport


mijicd:
At the moment we support reliable transport via TCP. Ideally, we should support the best effort, UDP, transport.

plokhotnyuk:
How about using of [Aeron](https://github.com/real-logic/aeron)?

mijicd:
@plokhotnyuk While Aeron is amazing, I believe at first we'll lean toward using `zio-nio`. However, with proper wrapper, Aeron might be a nice alternative.

pshemass:
@plokhotnyuk as @mijicd mentioned we try to use as `zio-nio` for the time being. We are aware of `Aeron`, `ZeroMQ`, `Chronicle` and couple of others but mission of this library is not providing wrappers for other libraries which you can create yourself. This library is to provide performant building blocks that can be use to build distributed system from the ground.

mijicd:
cc @vasilmkd

vasilmkd:
I'll be working on this. Thanks @mijicd.

pshemass:
implemented by #196

PR #178, opened 2020-02-21
UDP transport first try

vasilmkd:
WIP #133.

I left a comment in the `bind` method on the `Transport` service. At the moment, communicating over a datagram channel doesn't exactly correspond to the interface of the service. However, I'm in no way expert on this, so I might be wrong.

Help is very much appreciated.

### DIFF
--- a/keeper/src/main/scala/zio/keeper/transport/udp.scala
+++ b/keeper/src/main/scala/zio/keeper/transport/udp.scala
@@ -1,0 +1,137 @@
+package zio.keeper.transport
+
+import java.io.IOException
+
+import zio._
+import zio.clock.Clock
+import zio.duration.Duration
+import zio.keeper.TransportError
+import zio.keeper.TransportError._
+import zio.logging.Logging
+import zio.macros.delegate._
+import zio.nio.SocketAddress
+import zio.nio.channels.DatagramChannel
+
+object udp {
+
+  def withUdpTransport(
+    connectionTimeout: Duration,
+    requestTimeout: Duration,
+    chunkSize: Int
+  ) = enrichWithM[Transport](udpTransport(connectionTimeout, requestTimeout, chunkSize))
+
+  def udpTransport(
+    connectionTimeout: Duration,
+    requestTimeout: Duration,
+    chunkSize: Int
+  ): ZIO[Clock with Logging[String], Nothing, Transport] = {
+    val connectionTimeout_ = connectionTimeout
+    val requestTimeout_ = requestTimeout
+    val chunkSize_ = chunkSize
+    ZIO.environment[Clock with Logging[String]].map { env =>
+      new Live with Clock {
+        override val connectionTimeout: Duration = connectionTimeout_
+        override val requestTimeout: Duration = requestTimeout_
+        override val logger: Logging.Service[Any, String] = env.logging
+        override val clock: Clock.Service[Any] = env.clock
+        override val chunkSize: Int = chunkSize_
+      }
+    }
+  }
+
+  trait Live extends Transport { self: Clock =>
+    val connectionTimeout: Duration
+    val requestTimeout: Duration
+    val chunkSize: Int
+
+    val logger: Logging.Service[Any, String]
+
+    val transport = new Transport.Service[Any] {
+
+      override def connect(to: SocketAddress): ZManaged[Any, TransportError, ChannelOut] =
+        (for {
+          datagramChannelAndClose  <- DatagramChannel().withEarlyRelease.mapError(ExceptionWrapper)
+          (close, datagramChannel) = datagramChannelAndClose
+          _ <- datagramChannel
+                .connect(to)
+                .mapError(ExceptionWrapper)
+                .timeoutFail(ConnectionTimeout(to, connectionTimeout))(connectionTimeout)
+                .toManaged_
+          _ <- logger.info("transport connected to " + to).toManaged_
+        } yield new DatagramChannelOut(datagramChannel, to, requestTimeout, chunkSize, close, self))
+          .provide(self)
+
+      override def bind(
+        addr: SocketAddress
+      )(connectionHandler: ChannelOut => zio.UIO[Unit]): ZManaged[Any, TransportError, ChannelIn] =
+        // There is no way to use the datagram channel as a `ChannelOut` because
+        // it is not connected to any remote address (read on `ChannelOut` cannot be called).
+        // Also, UDP datagram channels are used with `receive`, instead of `read`
+        // because one does not need to necessarily care where the data comes from.
+        // Calling `read` on a connected datagram channel only allows receiving bytes from
+        // the remote address.
+        DatagramChannel()
+          .flatMap(dc => dc.bind(addr).toManaged_.as(dc))
+          .mapError(BindFailed(addr, _))
+          .withEarlyRelease
+          .onExit { _ =>
+            logger.info("shutting down udp channel")
+          }
+          .map {
+            case (close, datagramChannel) =>
+              new DatagramChannelIn(datagramChannel, close)
+          }
+    }
+  }
+
+  private class DatagramChannelOut(
+    datagramChannel: DatagramChannel,
+    remoteAddress: SocketAddress,
+    requestTimeout: Duration,
+    chunkSize: Int,
+    finalizer: URIO[Any, Any],
+    clock: Clock
+  ) extends ChannelOut {
+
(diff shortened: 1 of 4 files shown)


keeper/src/main/scala/zio/keeper/transport/udp.scala
```diff
@@ -0,0 +1,137 @@
...
+    override def isOpen: ZIO[Any, TransportError, Boolean] =
+      datagramChannel.isOpen
+
+    override def read: ZIO[Any, TransportError, Chunk[Byte]] =
+      datagramChannel
+        .read(chunkSize)
+        .map(_.fold[Chunk[Byte]](Chunk.empty)(identity))
+        .mapError(ExceptionWrapper)
+        .catchSome {
+          case ExceptionWrapper(ex: IOException) if ex.getMessage == "Connection reset by peer" =>
```
pshemass: do you think that will ever happen for UDP?

keeper/src/main/scala/zio/keeper/transport/udp.scala
```diff
@@ -0,0 +1,137 @@
...
+            close *> IO.fail(ExceptionWrapper(ex))
+        }
+
+    override def send(data: Chunk[Byte]): ZIO[Any, TransportError, Unit] =
+      datagramChannel
+        .write(data)
+        .retry(Schedule.recurs(3))
```
pshemass: this is kind of hack that I made but we should synchronize this with semaphore or something. I'm working on refactor of this https://github.com/pshemass/scalaz-distributed/blob/swim_refactor/keeper/src/main/scala/zio/keeper/transport/tcp.scala#L53

keeper/src/main/scala/zio/keeper/transport/udp.scala
```diff
@@ -0,0 +1,137 @@
...
+        .mapError(ExceptionWrapper)
+        .unit
+        .catchSome {
+          case ExceptionWrapper(ex: IOException) if ex.getMessage == "Connection reset by peer" =>
```
pshemass: again do you thing that this will exist?

keeper/src/main/scala/zio/keeper/transport/udp.scala
```diff
@@ -0,0 +1,137 @@
...
+object udp {
+
+  def withUdpTransport(
+    connectionTimeout: Duration,
+    requestTimeout: Duration,
+    chunkSize: Int
+  ) = enrichWithM[Transport](udpTransport(connectionTimeout, requestTimeout, chunkSize))
+
+  def udpTransport(
+    connectionTimeout: Duration,
```
pshemass: connection timeout and request timeout in udp?

keeper/src/main/scala/zio/keeper/transport/udp.scala
```diff
@@ -0,0 +1,137 @@
...
+          _ <- logger.info("transport connected to " + to).toManaged_
+        } yield new DatagramChannelOut(datagramChannel, to, requestTimeout, chunkSize, close, self))
+          .provide(self)
+
+      override def bind(
+        addr: SocketAddress
+      )(connectionHandler: ChannelOut => zio.UIO[Unit]): ZManaged[Any, TransportError, ChannelIn] =
+        // There is no way to use the datagram channel as a `ChannelOut` because
+        // it is not connected to any remote address (read on `ChannelOut` cannot be called).
+        // Also, UDP datagram channels are used with `receive`, instead of `read`
```
pshemass: what do you think about changing interface ?

```
trait ChannelOut {
   def send: (SocketAddress, Chunk[Byte]) => ZIO[Any, TransportError, Unit]
   def read: ZStream[Any, TransportError, (SocketAddress, Chunk[Byte])]
}

**Comment by pshemass** [comment]
@vasilmkd Looks pretty good. Could you as part of this PR also update version of `zio-nio` to latest ?

**Comment by vasilmkd** [comment]
@pshemass The PR is not great at the moment, but we'll fix it, you're right, I got carried away with the copying of the old code. Not everything is needed. It looks like we have to make changes to the interface. If you think that what you sketched out works, I will happily change it.

**Comment by pshemass** [comment]
@vasilmkd it was more question to understand what is your opinion. What do you think?

**Comment by mschuwalow** [comment]
I think it's in general still rather unclear whether we should unify TCP and UDP under one interface.

I think whatever we come up with will end up being a quite leaky abstraction. In our case this is especially misleading as a number of algorithms depend on TCP's ability to detect node failures, so they wouldn't work correctly with a UDP based transport unless we essentially reimplement TCP on top. I suggest we keep those two completely separate for now.

**Comment by pshemass** [comment]
@mschuwalow I agree with that because there is no nice way to create this abstraction :) We should probably improve `DatagramChannel` in `zio-nio` because it looks odd and method `bind` and `connect` are potentially leaky.

**Comment by vasilmkd** [comment]
I have already opened a PR to `zio-nio` for missing APIs in `DatagramChannel`. I also tweaked some signatures and implementations. https://github.com/zio/zio-nio/pull/155

**Comment by mijicd** [comment]
@vasilmkd thanks for submitting this PR, it's going to be a very valuable addition and it is definitely shaping up!

@mschuwalow abstracting over TCP and UDP is somewhat dangerous thing to do, so I'd definitely stay away from that.

@pshemass The only thing I'd do is move `close` to the part where transport is instantiated (as part of release).

File: keeper/src/test/scala/zio/keeper/transport/UdpSpec.scala [hunk 6]
```diff
@@ -0,0 +1,38 @@
...
+    testM("can send and receive messages") {
+      checkM(Gen.listOf(Gen.anyByte)) { bytes =>
+        val payload = Chunk.fromIterable(bytes)
+
+        environment >>> Live.live(for {
+          port         <- freePort
+          addr         <- SocketAddress.inetSocketAddress(port)
+          startPromise <- Promise.make[Nothing, Unit]
+          chunk        <- bindAndWaitForValue(addr, startPromise).fork
+          _            <- UIO.effectTotal(println("blah"))
```
mijicd: Please replace with `putStrLn`.

keeper/src/test/scala/zio/keeper/transport/TransportUtil.scala
```diff
@@ -0,0 +1,55 @@
...
+import zio.clock.Clock
+import zio.console._
+import zio.logging.Logging
+import zio.logging.slf4j.Slf4jLogger
+import zio.macros.delegate._
+import zio.macros.delegate.syntax._
+import zio.nio.SocketAddress
+import zio.test.environment.Live
+
+object TransportUtil {
```
mijicd: If possible, find a more specific place, utils usually end up being "trash bins" :).

keeper/src/main/scala/zio/keeper/transport/udp.scala
```diff
@@ -0,0 +1,137 @@
...
+    requestTimeout: Duration,
+    chunkSize: Int,
+    finalizer: URIO[Any, Any],
+    clock: Clock
+  ) extends ChannelOut {
+
+    override def isOpen: ZIO[Any, TransportError, Boolean] =
+      datagramChannel.isOpen
+
+    override def read: ZIO[Any, TransportError, Chunk[Byte]] =
```
mijicd: Save some typing, `IO[TransportError, Chunk[Byte]]`. Also, it can be `val` (the same applies for all other effect defs).

pshemass:
> @vasilmkd thanks for submitting this PR, it's going to be a very valuable addition and it is definitely shaping up!
> 
> @mschuwalow abstracting over TCP and UDP is somewhat dangerous thing to do, so I'd definitely stay away from that.
> 
> @pshemass The only thing I'd do is move `close` to the part where transport is instantiated (as part of release).

@mijicd code looks good over all and NioChannel is not good enough. We can merge this soon but  we should decide what to do with transport interface.

vasilmkd:
This is far from ready to be merged, it doesn't even work yet, and there are tests missing. I'm waiting for a decision on the interface, and this will need another serious review pass.

pshemass:
@vasilmkd your changes in `zio-nio` have been merged and released as 1.0.0-RC3

vasilmkd:
I will give it another go, thanks.

keeper/src/main/scala/zio/keeper/transport/udp.scala
mschuwalow: We should allign on which one we want to keep, as we currently have 2 tcp transport implementations

pshemass:
closing this in favor of #196

PR #188, opened 2020-03-07
ZIO 1.0.0-RC18-2 upgrade

vasilmkd:
Opening for purposes of progress tracking and early feedback.


pshemass:
@vasilmkd looks good.

vasilmkd:
Continuing my work, I was a bit busy with university these days.

vasilmkd:
I'd really appreciate some help with the hanging `membership` `TransportSpec`.

@mschuwalow @pshemass @mijicd

pshemass:
> I'd really appreciate some help with the hanging `membership` `TransportSpec`.
> 
> @mschuwalow @pshemass @mijicd

@mschuwalow could you help him ? I looked into that but there is nothing obvious.

mschuwalow:
@vasilmkd 
if you update the test environment to be
```scala
  val environment =
    (Clock.live ++ Logging.make((_, msg) => UIO.effectTotal(println(msg)))) >>> tcp.tcpTransport(
      10,
      100.millis,
      100.millis,
      10.millis
    ) ++ Clock.live
```
it works. Problem was that the test clock was being used for the retries -> never retried

mschuwalow:
I also noticed that the last test does not look the way it is supposed to. Could you update it to something like:
```scala
      testM("can receive messages after bind stream is closed") {
        checkM(Gen.listOf(Gen.anyByte)) { bytes =>
          val payload = Chunk.fromIterable(bytes)
          for {
            trans <- ZIO.environment[Transport[Address]].map(_.get)
            fiber <- trans
                      .bind(addr)
                      .take(1)
                      .runHead
                      .flatMap(c => ZStream.managed(c.get).flatMap(_.receive.take(1)).runHead)
                      .fork
            _      <- trans.send(addr, payload).retry(Schedule.spaced(10.milliseconds))
            result <- fiber.join
          } yield assert(result)(isSome(equalTo(payload)))
        }
      }
```

Original idea was to make sure that connections survive the end of the outer stream

vasilmkd:
Thanks, working on this tonight.

vasilmkd:
@adamgfraser Can I please ask you to take a look at the `ClassLoaderLayeringStrategy` errors that occur? This only started happening after I moved from `ZIO` `1.0.0-RC18` to `1.0.0-RC18-2`. Thanks. 🙏

adamgfraser:
@vasilmkd My hypothesis would be that it is related to zio/zio#3058. We changed `Tagged` there from being a case class that wrapped a type tag to a type alias for one. That was one of the last PRs that was merged before released 18-2 and it looks like there was another successful snapshot earlier that day (see [here](https://github.com/zio/zio/compare/v1.0.0-RC18-1...v1.0.0-RC18-2) so you could try testing with that to provide some support for this hypothesis.

Are you sure that all of your dependencies have been updated to 18-2? My guess would be that something has not been which is causing this error, but I could be wrong.

vasilmkd:
You're right, I'm pretty sure `zio-logging` hasn't been updated yet. I'll wait then. Thanks @adamgfraser.

vasilmkd:
The `Views[T]` layer needs to be hidden away, but cannot at the moment due to missing derivation of `Tagged`. We should be able to fix it when we further migrate to a newer RC.

vasilmkd:
@adamgfraser Can confirm, after `zio-logging` was upgraded to the latest `RC(2)`, the error went away. Thanks again.

adamgfraser:
@vasilmkd Great!

pshemass:
@vasilmkd great job here!

vasilmkd:
I don't anticipate much follow up here. I managed to hide the `Views[T]` layer with the latest snapshot of ZIO. So everything is back to normal. Thanks for everyone's help.

mijicd:
Love it!

PR #195, opened 2020-03-14
moved ByteCodec and TaggedCodec to core

pshemass:


keeper/src/main/scala/zio/keeper/errors.scala
```diff
@@ -15,16 +13,28 @@ sealed abstract class SerializationError(msg: String = "") extends Error(msg = m
 
 object SerializationError {
 
-  final case class SerializationTypeError[A](val cause: Throwable)(implicit ct: ClassTag[A])
+  final case class SerializationTypeError(msg0: String)
```
mschuwalow: we should probably also set the exception cause in order to preserve stack traces

keeper/src/main/scala/zio/keeper/membership/ByteCodec.scala
```diff
@@ -17,9 +19,9 @@ object ByteCodec {
     new ByteCodec[A] {
 
       def toChunk(a: A) =
-        ZIO.effect(Chunk.fromArray(writeBinary(a)(rw))).mapError(SerializationError("Failed serializing to chunk.", _))
+        ZIO.effect(Chunk.fromArray(writeBinary(a)(rw))).mapError(SerializationTypeError(_))
```
mschuwalow: I don't know whether calling this a TypeError is a good idea

pshemass: why?

pshemass: I will changed that

keeper/src/main/scala/zio/keeper/errors.scala
pshemass: I will add that

PR #203, opened 2020-03-20
Referentially transparent TRandom

vasilmkd:
@mschuwalow Would you be interested in something like this? It really bugs me that the current implementation uses an impure function.


mschuwalow:
I'm quite sure the current one is referentially transparent already. if you can show that it isn't I'll gladly merge this one.

PR #208, opened 2020-04-05
embrace zstm

mschuwalow:
No logic changes, just cleanup

### DIFF
--- a/membership/src/main/scala/zio/membership/hyparview/package.scala
+++ b/membership/src/main/scala/zio/membership/hyparview/package.scala
@@ -5,7 +5,7 @@ import java.util.UUID
 import zio._
 import zio.logging.Logging.Logging
 import zio.logging.log
-import zio.stm.STM
+import zio.stm.{ STM, ZSTM }
 import zio.stream._
 import zio.keeper.membership.{ ByteCodec, TaggedCodec }
 import zio.membership.hyparview.NeighborReply.{ Accept, Reject }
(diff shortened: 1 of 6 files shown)


membership/src/main/scala/zio/membership/hyparview/package.scala
```diff
@@ -132,75 +132,68 @@ package object hyparview {
...
-                                                ActiveProtocol
-                                                  .ForwardJoin(views.myself, msg.sender, TimeToLive(config.arwl))
-                                              )
-                                        )
-                                  reply <- ByteCodec
-                                            .toChunk(JoinReply(views.myself))
-                                            .mapError(e => zio.membership.SerializationError(e.msg))
-                                  _ <- con.send(reply)
-                                } yield Some(msg.sender)
+                                  full <- Views.isActiveViewFull
```
mijicd, line 156: We have `ifM` as well :).

membership/src/main/scala/zio/membership/hyparview/package.scala
```diff
@@ -228,17 +221,14 @@ package object hyparview {
...
-                    activeViewSize <- views.activeViewSize
-                    _              <- STM.check(activeViewSize < views.activeViewCapacity)
-                    nodeOpt        <- views.passiveView.flatMap(xs => tRandom.selectOne(xs.toList))
-                    node           <- nodeOpt.fold[STM[Nothing, T]](STM.retry)(STM.succeed(_))
-                  } yield (node, Neighbor(views.myself, activeViewSize <= 0))
-                }
-              }
+            ZSTM.atomically {
+              for {
+                _              <- Views.isActiveViewFull.flatMap(full => STM.check(!full))
```
mijicd, line 226: You can use `retryWhile` or `retryUntil` here.

membership/src/main/scala/zio/membership/hyparview/periodic.scala
```diff
@@ -9,48 +9,43 @@ import zio.logging.Logging.Logging
...
+                    node,
+                    ActiveProtocol
+                      .Shuffle(localAddr, localAddr, active, passive, TimeToLive(config.shuffleTTL))
+                  )
+                  .as(state)
+            }
+          }
+          .commit
+          .flatten
+      go.eventually
```
mijicd, line 36: You can remove val here :).

membership/src/main/scala/zio/membership/hyparview/periodic.scala
```diff
@@ -9,48 +9,43 @@ import zio.logging.Logging.Logging
...
-          go.eventually
-        }
-      }
+    getConfig.flatMap { config =>
+      val go: ZIO[TRandom with Views[T], SendError, ViewState] =
+        Views.activeView
+          .map(_.toList)
+          .flatMap { nodes =>
+            TRandom.selectOne(nodes).flatMap {
+              case None => STM.succeed(Views.viewState.commit)
```
mijicd, line 18: Why not `Views.viewState` only?

mschuwalow: It expects an STM[_, UIO[_]], so I rolled with this one

mschuwalow:
@mijicd thanks for your review. I think I addressed most of the points, can you take another look?

I'm also running into flakiness with the tests as they are currently using a fixed port for the TCPTransportSpec. Sadly I can't think of a good way to fix that without polluting our interfaces. Maybe you can take a look?

PR #213, opened 2020-04-11
[WIP] Leader Election "Coordinator" Module & JGroups Module 

mtsokol:
Hi @mijicd @pshemass @mschuwalow @vasilmkd !

Here's the **PoC** of "coordinator" leader election and JGroups module together with "Distributed HashMap" usage example. As I said it's the simplest possible approach I could think of.

I would be thankful if you could look at it in spare moment 🙂 

---

## Coordinator leader election

Each node has a list where it keeps order of joining nodes (**view**). Head of that list is a leader (coordinator). 
Then a new node is joining:

1. If it's first in cluster it just starts Consensus layer
2. If it's not the first one then it broadcasts `ViewRequest` and only leader is expected to response with `ViewResponse(view)`. Starting node will not start consensus layer until it receives `ViewResponse` (will periodically resend the request). After receiving it consensus layer is run.

Then a new node is joining other nodes will append theirs views, when a node leaves other nodes remove will remove it from view (**via membership events**). 

**When a leader leaves the cluster then it's successor will broadcast it's view to everyone.**

## JGroups module

It's functionality of maintaining a distributed state.
Then a new node is joining:

1. If it's first in cluster it just starts JGroups layer and use it's `initial state` to set the state
2. If it's not the first one then it asks Leader for a state and waits for it (will periodically resend the request). After setting the received state it starts JGroups layer.

Defined events will be broadcasted to all nodes. If node receives a `StateRequest` it checks whether is a leader and (if yes) it responds with `StateResponse`.

There is a `ReceiverAdapter` implemented by user defining State and behavior for receiving a message.  

## Distributed HashMap example

We have a `SimpleStringMap` that we want to have as a distributed one. We implement it together with `ReceiverAdapter` and also create `JGroups.live` to have access to `send` method. We compose it all together and use a ready data structure.

You can run these three Apps and send `PUT key value` or `REMOVE key`. Operations will be sent and new nodes will receive state from leader. 

---

Leaving a node seems broken still, haven't fixed it yet. 

As I said this is a **PoC**. 

What do you think about it?


pshemass:
@mtsokol this looks pretty good I will try to go through this next week

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
```diff
@@ -0,0 +1,172 @@
...
+import zio.stream.{ Stream, ZSink }
+import zio.keeper.protocol.InternalProtocol._
+import upickle.default._
+import zio.duration.Duration
+import zio.logging.Logging
+
+object Consensus {
+
+  trait Service {
+    def getLeader: UIO[NodeId]
```
pshemass, line 18: I think this service should 

```
trait Service {
   val leader: IO[NoLeaderError, NodeId]
   val onLeaderChange: Stream[Error, NodeId]
}
```

pshemass: rest of the methods are in ```Membership``` which will be required but consensus I believe

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
```diff
@@ -0,0 +1,172 @@
...
+        case 50 => viewRequestCodec.asInstanceOf[ByteCodec[ConsensusMsg]]
+        case 51 => viewResponseCodec.asInstanceOf[ByteCodec[ConsensusMsg]]
+        case 52 => userMsgCodec.asInstanceOf[ByteCodec[ConsensusMsg]]
+      }
+    )
+
+  private def runEventLoop(ref: Ref[List[NodeId]], self: NodeId) =
+    membership.events.foreach {
+      case MembershipEvent.Join(member) =>
+        ref.update(view => view :+ member.nodeId)
```
mijicd, line 62: This one's going to be inefficient, due to list's "performance" on appending. Alternatives:

- convert to prepend if you can live with reversed list
- use vector
- wait for chunk to support efficient append

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
```diff
@@ -0,0 +1,172 @@
...
+
+      case MembershipEvent.Leave(member) =>
+        for {
+          successor <- ref.modify { view =>
+                        val idx       = view.indexOf(member.nodeId)
+                        val successor = if (idx == 0) view.tail.headOption else None
+                        (successor, view.filterNot(_ == member.nodeId))
+                      }
+          _ <- successor match {
+                case Some(id) if id == self =>
```
mijicd, line 72: This is an incomplete pattern match.

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
```diff
@@ -0,0 +1,172 @@
...
+      _       <- logging.logInfo(s"[CONSENSUS] Asking for a view")
+      _       <- mem.broadcast(payload)
+      response = mem.receive
+        .mapM(msg => TaggedCodec.read[ConsensusMsg](msg.payload))
+        .collect { case s: ConsensusViewResponse => s }
+        .run(ZSink.head[ConsensusViewResponse])
+        .flatMap {
+          case None => UIO(false)
+          case Some(s) =>
+            logging.logInfo(s"[CONSENSUS] Received view: ${s.view}") *> ref.set(s.view).map(_ => true)
```
mijicd, line 131:
```suggestion
            logging.logInfo(s"[CONSENSUS] Received view: ${s.view}") *> ref.set(s.view).as(true)
```

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
```diff
@@ -0,0 +1,172 @@
...
+        }
+      timeout = clock.sleep(Duration.fromNanos(3000000000L)) *> UIO(false)
+      result  <- response.race(timeout)
+    } yield result
+
+  def create(): ZLayer[Membership with Clock with Logging, Error, Consensus] =
```
mijicd, line 137: `val live`?

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+    def setState(data: Chunk[Byte]): IO[Error, Unit]
+
+    def getState: IO[Error, Chunk[Byte]]
+  }
+
+}
+
+object JGroups {
+
+  sealed private trait JGroupsProtocol
```
mijicd, line 29: Same as above.

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+                              } yield ()
+                            } else UIO.unit
+                      } yield ()
+                    case StateResponse(payload) =>
+                      for {
+                        _      <- logging.logInfo(s"[JGROUPS] Received state response")
+                        leader <- c.getLeader
+                        _ <- if (leader == rawMsg.sender) {
+                              ra.setState(payload)
+                            } else UIO.unit
```
mijicd, line 103:
```suggestion
                        _ <- UIO.when(leader == rawMsg.sender)(ra.setState(payload))
```

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+                        _      <- logging.logInfo(s"[JGROUPS] Received state request")
+                        leader <- c.getLeader
+                        _ <- if (leader == selfId) {
+                              for {
+                                state   <- ra.getState
+                                payload <- TaggedCodec.write[JGroupsProtocol](StateResponse(state))
+                                _       <- c.send(payload, rawMsg.sender)
+                                _       <- logging.logInfo(s"[JGROUPS] Answering state request")
+                              } yield ()
+                            } else UIO.unit
```
mijicd, line 95:
```suggestion
                        leader <- c.getLeader
                        _ <- UIO.whenM(c.getLeader.map(_ == selfId)) {
                              for {
                                state   <- ra.getState
                                payload <- TaggedCodec.write[JGroupsProtocol](StateResponse(state))
                                _       <- c.send(payload, rawMsg.sender)
                                _       <- logging.logInfo(s"[JGROUPS] Answering state request")
                              } yield ()
                            } 
```

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+
+  private val msgLoop: ZIO[Consensus with ReceiverAdapter with Logging, Error, Unit] =
+    for {
+      c      <- ZIO.access[Consensus](_.get[Consensus.Service])
+      ra     <- ZIO.access[ReceiverAdapter](_.get[ReceiverAdapter.Service])
+      selfId <- c.selfNode
+      _ <- c.receive.tap { rawMsg =>
+            for {
```
mijicd, line 81: Nitpick: Dropping this for-comprehension in favor of plain `flatMap` would increase readability here.

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+            payload <- TaggedCodec.write[JGroupsProtocol](UserMessage(data))
+            _       <- c.broadcast(payload)
+          } yield ()
+      }
+    }
+
+  private val msgLoop: ZIO[Consensus with ReceiverAdapter with Logging, Error, Unit] =
+    for {
+      c      <- ZIO.access[Consensus](_.get[Consensus.Service])
+      ra     <- ZIO.access[ReceiverAdapter](_.get[ReceiverAdapter.Service])
```
mijicd, line 78: Just take the whole env.

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  val live: ZLayer[Consensus, Error, JGroups] =
+    ZLayer.fromService[Consensus.Service, JGroups.Service] { c =>
+      new Service {
+        override def send(data: Chunk[Byte]): IO[Error, Unit] =
+          for {
+            payload <- TaggedCodec.write[JGroupsProtocol](UserMessage(data))
+            _       <- c.broadcast(payload)
+          } yield ()
```
mijicd, line 71:
```suggestion
          TaggedCodec.write[JGroupsProtocol](UserMessage(data)).flatMap(c.broadcast)
```

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
```diff
@@ -0,0 +1,172 @@
...
+      response = mem.receive
+        .mapM(msg => TaggedCodec.read[ConsensusMsg](msg.payload))
+        .collect { case s: ConsensusViewResponse => s }
+        .run(ZSink.head[ConsensusViewResponse])
+        .flatMap {
+          case None => UIO(false)
+          case Some(s) =>
+            logging.logInfo(s"[CONSENSUS] Received view: ${s.view}") *> ref.set(s.view).map(_ => true)
+        }
+      timeout = clock.sleep(Duration.fromNanos(3000000000L)) *> UIO(false)
```
mijicd, line 133:
```suggestion
      timeout = clock.sleep(3.seconds) *> UIO(false)
```

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
```diff
@@ -0,0 +1,172 @@
...
+  def create(): ZLayer[Membership with Clock with Logging, Error, Consensus] =
+    ZLayer.fromEffect {
+      for {
+        queue       <- Queue.unbounded[Message]
+        selfId      <- membership.localMember
+        _           <- logging.logInfo(s"[CONSENSUS] Starting consensus layer for node: ${selfId.nodeId}")
+        members     <- membership.nodes
+        _           <- logging.logInfo(s"[CONSENSUS] Present nodes: $members")
+        consMembers <- Ref.make(List(selfId.nodeId))
+        _           <- if (members.nonEmpty) setCurrentView(consMembers).doUntilEquals(true) else UIO.unit
```
mijicd, line 146:
```suggestion
        _           <- UIO.when(members.nonEmpty)(setCurrentView(consMembers).doUntilEquals(true))
```

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+      _       <- logging.logInfo(s"[JGROUPS] Asking for a state")
+      _       <- c.send(payload, leader)
+      response = c.receive
+        .filter(_.sender == leader)
+        .mapM(msg => TaggedCodec.read[JGroupsProtocol](msg.payload))
+        .collect { case s: StateResponse => s }
+        .run(ZSink.head[StateResponse])
+        .flatMap {
+          case None    => UIO(false)
+          case Some(s) => ra.setState(s.payload).map(_ => true)
```
mijicd, line 127:
```suggestion
          case Some(s) => ra.setState(s.payload).as(true)
```

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+        }
+      timeout = clock.sleep(Duration.fromNanos(3000000000L)) *> UIO(false)
```
mijicd, line 129:
```suggestion
      timeout = clock.sleep(3.seconds) *> UIO(false)
```

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+      result  <- response.race(timeout)
+    } yield result
+
+  def create[T: Tagged](): ZLayer[Consensus with ReceiverAdapter with Clock with Logging with Has[T], Error, Has[T]] =
+    ZLayer.fromEffect {
+      for {
+        _   <- clock.sleep(Duration.fromNanos(3000000000L))
```
mijicd, line 136:
```suggestion
        _   <- clock.sleep(3.seconds)
```

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+        .run(ZSink.head[StateResponse])
+        .flatMap {
+          case None    => UIO(false)
+          case Some(s) => ra.setState(s.payload).map(_ => true)
+        }
+      timeout = clock.sleep(Duration.fromNanos(3000000000L)) *> UIO(false)
+      result  <- response.race(timeout)
+    } yield result
+
+  def create[T: Tagged](): ZLayer[Consensus with ReceiverAdapter with Clock with Logging with Has[T], Error, Has[T]] =
```
mijicd, line 133: No need for `()`, this is a pure function. *Nitpick:* start from `A` with type params.

keeper/src/main/scala/zio/keeper/consensus/JGroups.scala
```diff
@@ -0,0 +1,147 @@
...
+    ZLayer.fromEffect {
+      for {
+        _   <- clock.sleep(Duration.fromNanos(3000000000L))
+        env <- ZIO.environment[Consensus with ReceiverAdapter with Logging with Has[T]]
+        t   <- ZIO.access[Has[T]](_.get[T])
+        c   = env.get[Consensus.Service]
+        v   <- c.getView
+        _   <- if (v.size > 1) initializeState.doUntilEquals(true) else UIO.unit
```
mijicd, line 141:
```suggestion
        _   <- UIO.when(v.size > 1)(initializeState.doUntilEquals(true))
```

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
mtsokol: OK, so then as I understand we have access to all modules from our protocol stack? Not only to the one on the top?

mtsokol: Then I also agree

mtsokol: Last thing: `getView` is basically an ordered `nodes` from Membership according to the joining time. Remove it also?

mtsokol: ping

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
mtsokol: Sure, moved to Vector

mtsokol:
@mijicd @pshemass Hi, I've done changes according to first round of review + did a rebase and moved to `Membership[A]` + I've got still one question at the top.

So now joining, broadcasting events and sharing state by leader with new nodes works.

What is left:
- [ ] leaving consensus is still a mess... need to polish it to make sure changing leader and views occurs correctly
- [ ] test suites for consensus and jgroups
- [ ] docs

---

@pshemass I've also got three questions regarding changes I had to do while rebasing to the SWIM after refactor:
1. When starting SWIM it used to run after determining members. Now I had to slept starting higher layer as SWIM starts immediately with empty members list and updates it asynchronously. Did I get it right?
2. Had to switched from `broadcast` to `foreach send` as for messages sent via broadcast I was getting `No codec found for tag...` when receiving them, like it wasn't using provided `user` protocol and its `TaggedCodec[B]` in `Membership[B]`. Is that correct?
3. As I'm going to focus right now on leaving consensus I just wanted to ask if leaving cluster in current version of SWIM is all operational?

pshemass:
@mtsokol Now in Swim implementation list of nodes from discovery is taken and we send request to join to each of them. we add to list of `Healthy` nodes when we receive `Accept`.

this broadcast thing sounds like a bug. let me write test for this.

I believe you will get notification death node leaving cluster - nodes that are unreachable and decided to be removed by cluster. However `Leave` message is not implemented this should be trigger when graceful shutdown happen and you send `Death` message about yourself.

keeper/src/main/scala/zio/keeper/consensus/Consensus.scala
pshemass: I was thinking about this recently because I read some paper about leader election and what they did to achieve order they use IP addresses so 1.1.1.1 < 1.1.1.2. Maybe we should do similar thing because time in distributed systems is always a bit sneaky because clock can be skew or nodes join in the exactly same time. @mijicd what do you think?

mijicd: Yes, it would be great if we can avoid ordering anything by (wall clock) timestamps.

mtsokol: @pshemass Thanks! My side note: So here (for convergence) when a new leader is elected (because previous left) the new leader will broadcast it's view (ordered nodes) which other will always accept even if they would see different node as a leader, or had some messed up order of joined nodes, so (but in some edge cases this approach also might fail I guess). WDYT?

mtsokol: @pshemass sure, so as I understand election will be resolved by choosing smallest value of `address` + `port` value among members?

pshemass:
@mtsokol broadcast with user protocol is broken. working on fix. it should be ready by tomorrow.

mtsokol:
@mijicd @pshemass Hi! Please don't consider this PR as dead 😄. The end of remote semester took all my attention, but by the end of July I should get back to it.

I will rebase it then so if you have some remarks regarding general design / change of approach of this layer then let me know. (e.g. should it stay given way (nodes order by join time) or move it to fixed port ordering described by @pshemass?)

PR #237, opened 2020-05-19
plumtree tree optimization

### LINKED ISSUE #189: Plumtree tree optimization
I'm leaving this out of #180 for now.

Should be quite easy to do, but we need to decide what policy to adopt. There is one recommended in the paper, but I've seen projects use other, more sophisticated ones

mschuwalow:
closes #189

Followup will be done in seperate PRs.

### DIFF
--- a/keeper/src/main/scala/zio/keeper/membership/hyparview/ActiveProtocol.scala
+++ b/keeper/src/main/scala/zio/keeper/membership/hyparview/ActiveProtocol.scala
@@ -71,7 +71,7 @@ object ActiveProtocol {
   }
 
   final case class IHave(
-    uuids: ::[UUID]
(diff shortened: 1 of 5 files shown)


pshemass:
@mschuwalow could you elaborate a little bit what is all about because it's hard to review without context?

mschuwalow:
This is leftover work from the plumtree implementation. Basic idea is that if a messages is received multiple times via broadcast, the spanning tree might be optimized if the number of jumps can be reduced by more than some threshold.

Here the implementation of sendGraft is changed so that when we receive a message we check all remaining nodes that told us that the message is available via IHave messages. If one was more than the threshold faster we send a Graft message (can be optimized to use a new message type) which will cause it to transmit the messages and add our node to its eager push peers. This leads to the next broadcast being delivered faster

PR #238, opened 2020-05-23
Nodes layer

pshemass:
- I converted `Nodes` to layer
- introduced `ConversationId` that generates id for each message 
- organized messages and added test for serialization
- tests added for FailureDetection


keeper/src/main/scala/zio/keeper/errors.scala
```diff
@@ -4,7 +4,7 @@ import zio.duration.Duration
 import zio.keeper.transport.Address
 import zio.nio.core.SocketAddress
 
-sealed abstract class Error(val msg: String = "") {
+sealed abstract class Error(val msg: String = "") extends Throwable {
```
mijicd, line 7: I'd extend `Exception` (or `RuntimeException`). Not sure whether or not it make sense to disable stack trace accumulation, e.g. `extends Exception(msg, null, false, false)`.

keeper/src/main/scala/zio/keeper/membership/swim/ConversationId.scala
```diff
@@ -0,0 +1,32 @@
+package zio.keeper.membership.swim
+
+import zio.UIO
+import zio.ZLayer
+import zio.Ref
+import zio.ULayer
+import zio.ZIO
+import zio.URIO
```
mijicd, line 8:
```suggestion
import zio._
```

keeper/src/main/scala/zio/keeper/membership/swim/Nodes.scala
```diff
@@ -5,121 +5,89 @@ import zio.clock.Clock
...
+     * Node state for given NodeId.
+     */
+    def nodeState(id: NodeAddress): IO[Error, NodeState]
+
+    def numberOfNodes: UIO[Int]
+
+    /**
+     * Lists members that are in healthy state.
+     */
+    def healthyNodes: UIO[List[(NodeAddress, NodeState)]]
```
mijicd, line 60: Does it make sense to change to `Chunk`? Long-term goal in core is to migrate wherever possible. No need to do this now, just asking.

keeper/src/main/scala/zio/keeper/membership/swim/Nodes.scala
```diff
@@ -5,121 +5,89 @@ import zio.clock.Clock
...
+     * Returns next Healthy node.
+     */
+    val next: UIO[Option[NodeAddress]]
+
+    /**
+     * Node state for given NodeId.
+     */
+    def nodeState(id: NodeAddress): IO[Error, NodeState]
+
+    def numberOfNodes: UIO[Int]
```
mijicd, line 55:
```suggestion
    val numberOfNodes: UIO[Int]
```

keeper/src/main/scala/zio/keeper/membership/swim/Nodes.scala
pshemass: do you mean migrate ```List``` to ```Chunk```?

keeper/src/main/scala/zio/keeper/errors.scala
pshemass: @mschuwalow what do you think ?

mschuwalow: I'm personally a fan of keeping stack traces. Main reason for dropping them would be performance, no? In that case let's keep them for now and rip them out when we do a performance pass

keeper/src/main/scala/zio/keeper/membership/swim/Nodes.scala
mschuwalow: Yeah, I think we should start defaulting to Chunk.

keeper/src/test/scala/zio/keeper/membership/SwimSpec.scala
```diff
@@ -122,7 +123,7 @@ object SwimSpec {
               )
             )
           )
-      }.provideLayer(Clock.live ++ logging ++ (logging >>> TestDiscovery.live))
-    )
+      }.provideLayer(Clock.live ++ logging ++ (logging ++ Random.live >>> TestDiscovery.live))
+    ) @@ TestAspect.ignore
```
mschuwalow, line 127: What happened here?

pshemass: this is the test that is hanging on CI. locally I can run it with `nonFlaky` aspect but on CI it hangs :(

keeper/src/main/scala/zio/keeper/membership/swim/Nodes.scala
pshemass: let me try it.

keeper/src/main/scala/zio/keeper/errors.scala
pshemass: so I will change that to `Exception` and keep stacktrace for the time being.

pshemass: done

keeper/src/main/scala/zio/keeper/membership/swim/Nodes.scala
pshemass: let do this when `TMap` will have `toChunk` method.

mijicd: It's already there as of RC19.

pshemass: let do this in new PR

PR #244, opened 2020-05-29
K8 experiment - fixes

pshemass:
Couple of fixes after first run on Kubernetes.


mschuwalow:
Do we want to wait with this until @mijicd's PR is ready?

mijicd:
Let's merge this, one more adjustment won't change things for me.

build.sbt
```diff
@@ -63,9 +63,18 @@ lazy val keeper = project
     fork in test := true
   )
 
+lazy val dockerSettings = Seq(
+  dockerBaseImage := "openjdk:11",
```
mijicd, line 67: We can probably go with something thinner here, like one of those alpine jre images.

pshemass: sure

PR #246, opened 2020-05-29
Reorganize library

mijicd:
### Summary

- Removed "extra" methods from membership service.
- Defined keeper API.
- Flattened package hierarchy.
- Temporary removed SwimSpec as it depends on removed method and need to be refactored.


keeper/src/main/scala/zio/keeper/swim/Swim.scala
```diff
@@ -0,0 +1,74 @@
...
+        broadcast0 <- Broadcast.make(swimConfig.messageSizeLimit, swimConfig.broadcastResent).toManaged_
+        messages0  <- Messages.make(localNodeAddress, broadcast0, udpTransport)
+        _          <- messages0.process(swim).toManaged_
+      } yield new Membership.Service[B] {
+
+        override def broadcast(data: B): UIO[Unit] =
+          (for {
+            bytes <- ByteCodec.encode[User[B]](User(data))
+            _     <- broadcast0.add(Message.Broadcast(bytes))
+          } yield ()).orDie
```
pshemass, line 61: I'm not sure about that

PR #251, opened 2020-06-13
Lifeguard

pshemass:
Implementation of Lifeguard from https://arxiv.org/pdf/1707.00788.pdf
- Local Health Aware Probe (LHA-Probe) replaces the probing stage of SWIM’s failure detector, which has a fixed probe period and timeout, with one where they are dynamically adjusted, based on that member’s recent fail- ure detection-related communication with other members.
- Local Health Aware Suspicion (LHA-Suspicion) re- places the suspicion stage of SWIM’s failure detector, which has fixed suspicion timeouts, with one that has dynamic timeouts. The timeout for each new suspicion starts significantly higher than it would in the fixed case, but is reduced as independent suspicions about the same suspected member are processed.
- Buddy System replaces SWIM’s piggyback message selector with one that prioritizes notifying a suspected member of the suspicion, to reduce the average time to refutation, which helps both the Local Health Aware Probe and Local Health Aware Suspicion components be even more effective.


keeper/src/main/scala/zio/keeper/swim/protocols/FailureDetection.scala
```diff
@@ -40,78 +40,138 @@ object FailureDetection {
...
-                  case Some(Some((node, originalAckId))) =>
-                    ZIO.succeedNow(Message.Direct(node, originalAckId, Ack))
-                  case _ =>
-                    Message.noResponse
-                }
-            case Message.Direct(sender, conversationId, Ping) =>
-              ZIO.succeedNow(Message.Direct(sender, conversationId, Ack))
+    TMap
+      .empty[Long, Option[(NodeAddress, Long)]]
+      .zip(TMap.empty[Long, Unit])
```
mijicd, line 45: This is a TSet.

keeper/src/main/scala/zio/keeper/swim/protocols/FailureDetection.scala
```diff
@@ -3,22 +3,33 @@ package zio.keeper.swim.protocols
...
 
 object FailureDetection {
 
-  final case object Ping                        extends FailureDetection
-  final case object Ack                         extends FailureDetection
-  final case object Nack                        extends FailureDetection
-  final case class PingReq(target: NodeAddress) extends FailureDetection
+  final case object Ping                                           extends FailureDetection
+  final case object Ack                                            extends FailureDetection
+  final case object Nack                                           extends FailureDetection
```
mijicd, line 28: Drop final in front of these to avoid future compat issues.

keeper/src/main/scala/zio/keeper/swim/protocols/FailureDetection.scala
pshemass: done

keeper/src/main/scala/zio/keeper/swim/protocols/FailureDetection.scala
pshemass: done

PR #306, opened 2020-10-25
HyParView: no unnecessary messages & disconnect flow 

mschuwalow:
* Removes `JoinReply` as it is unnecessary for our flow
* Make sure `Disconnect` is only ever sent once and only from one node
* Introduce `ViewMessage` to make sure only `whitelisted` messages are sent to other nodes. Biggest offender is `Disconnect` and could introduce edge cases otherwise
* Move Neighbor sub-protocol to `Views`


keeper/src/main/scala/zio/keeper/hyparview/protocols.scala
```diff
@@ -5,35 +5,86 @@ import zio.stm._
...
+                      Views.doNeighbor.commit
+                  }
+            } yield ()
+          }
+          .onError(_ => Views.doNeighbor.commit)
+      case _: Message.Join =>
+        active
+      case _: Message.ForwardJoinReply =>
+        active
+      case _ =>
```
mschuwalow, line 76: We could introduce another `Message` subtype to remove this case, but I don't think it's worth the complexity.

keeper/src/main/scala/zio/keeper/hyparview/Views.scala
```diff
@@ -79,14 +101,54 @@ object Views {
       passiveViewCapacity <- passiveViewCapacity
     } yield ViewState(activeViewSize, activeViewCapacity, passiveViewSize, passiveViewCapacity)
 
-  def live: ZLayer[TRandom with HyParViewConfig, Nothing, Views] =
-    ZLayer.fromEffect {
+  def live: ZLayer[TRandom with HyParViewConfig with Clock, Nothing, Views] = {
+    def startNeighborHandler(
```
mschuwalow, line 105: This is implemented using an actor as we want to execute side effects to get the current time inside of the implementation.
Semantics should not break if we inline this into the service and execute inside of a transaction. Alternatively it would be nice to come up with a schema that does not need the current time

keeper/src/main/scala/zio/keeper/hyparview/protocols.scala
```diff
@@ -5,35 +5,93 @@ import zio.stm._
...
+    }
+
+  /**
+   * Start protocol with remote node.
+   * The protocol will be chosen depending on the passed message.
+   * If a neighbor message is sent, assume that this
+   * process is triggered by Views. In this case
+   * start a new doNeighbor process on failure.
+   */
+  def connectRemote(
```
mschuwalow, line 44: It would be nice to avoid introspecting the message here, but I can't think of a nice way that does not result in unnecessary message overhead.

keeper/src/main/scala/zio/keeper/hyparview/Message.scala
```diff
@@ -78,114 +100,85 @@ object Message {
...
-  }
-
-  final case class ForwardJoinReply(
-    sender: NodeAddress
-  ) extends Message
+  final case class Shuffle(
+    sender: NodeAddress,
+    originalSender: NodeAddress,
+    activeNodes: List[NodeAddress],
+    passiveNodes: List[NodeAddress],
```
mijicd, line 107: Can they be sets?

keeper/src/main/scala/zio/keeper/hyparview/Message.scala
```diff
@@ -78,114 +100,85 @@ object Message {
...
-
-  object JoinReply {
+  /**
+   * Messages used by plumtree layer.
+   */
+  sealed abstract class PeerMessage extends ViewMessage
 
-    implicit val codec: ByteCodec[JoinReply] =
-      ByteCodec.fromReadWriter(macroRW[JoinReply])
+  case object Prune extends PeerMessage {
```
mijicd, line 133: Let's put terms in companion.

keeper/src/main/scala/zio/keeper/hyparview/Views.scala
```diff
@@ -97,18 +159,36 @@ object Views {
...
         override def addAllToPassiveView(remaining: List[NodeAddress]): USTM[Unit] =
           remaining match {
             case Nil     => STM.unit
             case x :: xs => addToPassiveView(x) *> addAllToPassiveView(xs)
           }
 
-        override def addToActiveView(node: NodeAddress, send: Message => USTM[_], disconnect: USTM[_]): USTM[Unit] =
+        override def addShuffledNodes(sentOriginally: Set[NodeAddress], replied: Set[NodeAddress]): USTM[Unit] =
+          for {
+            _         <- STM.foreach(sentOriginally.toList)(removeFromPassiveView)
```
mijicd, line 173: Replace with `foreach_`.

keeper/src/test/scala/zio/keeper/hyparview/InitialProtocolSpec.scala
```diff
@@ -88,7 +89,7 @@ object InitialProtocolSpec extends KeeperSpec {
               makeConnection
                 .use { con =>
                   for {
-                    _           <- Views.addToActiveView(existingAddress, _ => STM.unit, STM.unit).ignore.commit
+                    _           <- Views.addToActiveView(existingAddress, _ => STM.unit, _ => STM.unit).ignore.commit
```
mijicd, line 92: You can factor out this lambda for slight efficiency benefit.

mijicd (commented):
Just a few minor comments. Make it pass and it's good to go.

keeper/src/main/scala/zio/keeper/hyparview/Message.scala
mschuwalow: There is trait hierarchy here that encodes what can be done with different subsets of messages. They already live in the Message companion object, but nesting them further makes usability really bad.

i.e. `Message.PeerMessage.Prune`

keeper/src/main/scala/zio/keeper/hyparview/Message.scala
mschuwalow: Yep, updated :+1:

mijicd (approved):

